PR #1, opened 2021-02-12
clean up react warnings; add a handful of unit tests

orzechowskid:
:wave: hi!  not sure what your plans are for unit testing, so I didn't go too wild in this PR, but here's a few dealing with some happy and sad paths through `<CovidAppointmentTable>` and cleaned up some React console warnings along the way.  let me know what you think!


test/fixtures/api/no-data.json
```diff
@@ -0,0 +1 @@
+{"statusCode":404,"body":"{}"}
```
orzechowskid, line 1: I haven't actually seen the /prod API call fail, so I had to take a guess here as to what it returns if no testing locations have available appointments.

johnhawkinson: Based on https://github.com/livgust/covid-vaccine-scrapers/blob/4210545b850182f573c47fd97248e223ecce5895/main.js#L61-L63

```js
		const responseJson = {
			results: finalResultsArray,
		};
```


and https://github.com/livgust/covid-vaccine-scrapers/blob/4210545b850182f573c47fd97248e223ecce5895/main.js#L80-L88

```js
			const params = {
				Bucket: process.env.AWSS3BUCKETNAME,
				Key: "data.json",
				Body: JSON.stringify(responseJson),
			};

			// Uploading files to the bucket
			return await new Promise((resolve, reject) => {
				s3.upload(params, function (err, data) {
```

at a minimum it'll have inside the body a null `results` array.
I think the better test is to count the length of the results array rather than testing for an explicit value.

orzechowskid: oh nice, thanks for the links to that repo.  I'll update the fixture to reflect that.

> I think the better test is to count the length of the results array rather than testing for an explicit value.

the length of the results array is just the length of the [mock data used by the unit test](https://github.com/livgust/macovidvaccines.com/pull/1/files#diff-49804f06dcb661d866cd1ae6a0494cb7d99cbe4b2019b149da6a7e0f7b52ccf4R38); testing against that is equivalent to testing that the mock function I wrote is doing the right thing.  the better test would be to count the number of rows rendered by the table but I can't do that since there is currently nothing in the rendered markup for [testing-library](https://testing-library.com/docs/dom-testing-library/api) to hook into (also, no table at all is rendered in the case of API call failure so there's nothing for the test framework to assert against).

I thought about adding those hooks since it would be pretty straightforward and if done right would provide a better experience for more user agents but I was hesitant to parachute into someone else's project and start changing too many things.  (for background, I offered to help with writing tests after a brief discussion with Olivia on Twitter the other day)

johnhawkinson: > the length of the results array is just the length of the mock data used by the unit test; testing against that is equivalent to testing that the mock function I wrote is doing the right thing.

Then I suspect we don't need the test at all.

src/CovidAppointmentTable.test.js
```diff
@@ -0,0 +1,73 @@
...
+
+    test('it correctly displays available appointments', async function() {
+      await act(async function() {
+        render(
+          <CovidAppointmentTable />
+        );
+      });
+
+      expect(await screen.findByText(/Berkshire Community College Field House/)).toBeInTheDocument();
+    });
```
livgust, line 33: I think that either this wording should be updated or the expect should be more generic. Making sure that one line of text appears isn't really testing that it displays correctly.

I recommend either:
1. Adding additional `expect`s that give a fuller picture of what we expect to be displayed, or
2. Renaming the test to something like 'it displays the name of a vaccine location' or something like that.

src/CovidAppointmentTable.test.js
```diff
@@ -0,0 +1,73 @@
...
+
+  describe('when no api data is available', function() {
+    beforeEach(function() {
+      window.fetch.mockResolvedValueOnce({
+        json: async () => noData,
+        ok: true
+      });
+    });
+
+    test('it displays a loader component', async function() {
```
livgust, line 44: I'm a little unclear here - are we testing that there's a "forever loader" essentially?

src/CovidAppointmentTable.test.js
```diff
@@ -0,0 +1,73 @@
...
+      await act(async function() {
+        render(
+          <CovidAppointmentTable />
+        );
+      });
+
+      // react-loader doesn't appear to expose any a11y attributes on the spinner it
+      // renders, which means we can't use the testing-library's API to assert on the
+      // presence of something; we have to assert on the absence of everything :(
+      expect(screen.queryByText(/Only show locations with available appointments/)).toBeNull();
```
livgust, line 54: Can we just check that the loader's `loaded` property is false? I think that's a little clearer and less fragile. I know some might not like that we're essentially trusting the react-loader component to show a spinner, but I think that we can rely on that component to do as it's told.

src/CovidAppointmentTable.test.js
```diff
@@ -0,0 +1,73 @@
...
+    });
+  });
+
+  describe('when the api endpoint can not be reached', function() {
+    beforeEach(function() {
+      window.fetch.mockImplementationOnce(() => Promise.reject(new TypeError('network error')));
+    });
+
+    test('it displays a loader component', async function() {
```
livgust, line 63: same comments here as in above test.

src/components/Availability.js
```diff
@@ -1,5 +1,4 @@
 export default function Availability({ entry }) {
-	console.log(entry.hasAppointments);
```
livgust, line 2: thank youuuuuu

test/fixtures/api/prod.json
```diff
@@ -0,0 +1 @@
...s\":0,\"hasAvailability\":false},\"3/8/2021\":{\"numberAvailableAppointments\":0,\"hasAvailability\":false},\"3/9/2021\":{\"numberAvailableAppointments\":0,\"hasAvailability\":false},\"3/10/2021\":{\"numberAvailableAppointments\":0,\"hasAvailability\":false},\"3/11/2021\":{\"numberAvailableAppointments\":0,\"hasAvailability\":false},\"3/12/2021\":{\"numberAvailableAppointments\":0,\"hasAvailability\":false},\"3/15/2021\":{\"numberAvailableAppointments\":1,\"hasAvailability\":true},\"3/16/2021\":{\"numberAvailableAppointments\":0,\"hasAvailability\":false},\"3/17/2021\":{\"numberAvailableAppointments\":0,\"hasAvailability\":false},\"3/18/2021\":{\"numberAvailableAppointments\":0,\"hasAvailability\":false},\"3/19/2021\":{\"numberAvailableAppointments\":0,\"hasAvailability\":false},\"3/22/2021\":{\"numberAvailableAppointments\":0,\"hasAvailability\":false},\"3/23/2021\":{\"numberAvailableAppointments\":0,\"hasAvailability\":false},\"3/24/2021\":{\"numberAvailableAppointments\":0,\"hasAvailability\":false},\"3/25/2021\":{\"numberAvailableAppointments\":0,\"hasAvailability\":false},\"3/26/2021\":{\"numberAvailableAppointments\":0,\"hasAvailability\":false}},\"hasAvailability\":true}]}"}
```
livgust, line 1: I think it would be more useful for testing purposes (and readability) if we formatted this like a regular JSON file and cut the results down to fewer locations. Maybe one with no availability, one that has multiple sign-up links with a drop-down (you can find an example of this when you see a signUpLink for each date of availability), one with multiple sign-up links but only one date available, and one with a single signUpLink.

livgust (changes requested):
orzechowskid - thank you so much for starting some unit tests here! 7 comments in total, LMK what you think. Also (congrats?) this is my first code review I've ever done on GitHub. So also LMK if I do something wrong hah!

orzechowskid:
> orzechowskid - thank you so much for starting some unit tests here! 7 comments in total, LMK what you think. Also (congrats?) this is my first code review I've ever done on GitHub. So also LMK if I do something wrong hah!

thanks for taking a look!  to be honest I'm not particularly happy with these tests - they would not pass muster at my day job, that's for sure - but I didn't want to just swoop into someone else's codebase and propose too many changes out of the blue.  so I limited myself to writing tests which, to the best of their ability, validated the existing behavior of the existing code.

if you're okay with me adding a few attributes to the HTML elements rendered by `<CovidAppointmentTable>` and its child components, and possibly changing some HTML tag names, then it would be super easy to write more descriptive and resilient tests which I think would take care of the (absolutely valid) questions you raised.

test/fixtures/api/prod.json
orzechowskid: I can remove some of the appointment entries from the results array, no problem.

but this file is already formatted as a regular JSON file; its contents are exactly what the browser receives when a request to the `/prod` API endpoint is made.  if you're specifically referring to the string of JSON stored in the `body` field, then I think there's value in keeping mock data identical to what's returned by a real API call but if you want I can store a decoded version of that string here in this file and re-encode it inside the unit tests (since the code is expecting a string, not an object).

livgust: That would be awesome if you could re-encode it like you suggested! That way we can add some indentation and new lines for readability which would be A+

livgust:
Feel free to change the source code - whatever you need! I'll let you know if it's too far for some reason, but I doubt it. Thanks again :)

src/CovidAppointmentTable.js
```diff
@@ -79,20 +88,34 @@ export default function CovidAppointmentTable() {
...
+        {ready && formattedData.length === 0 && (
+          <div role="status">
+            <p>No appointments found.</p>
+          </div>
+        )}
+            <div role="list">
 			{formattedData.map((entry) => {
 				return (
-					<div className={classes.cardBox}>
+				  	<div role="listitem" key={`${entry.location}-${entry.city}`} className={classes.cardBox}>
```
livgust, line 118: bad news - we have at least one instance (2 Price Choppers in Worcester) that make this key not unique :(

livgust (changes requested):
Thank you so much, this looks great. 3 nitpicky things:
1) left a comment re: key not being unique
2) could you install prettier and run it on these files? we have a config file in the repo now
3) conflicts due to other merged PRs, apologies!!!

LMK if you're swamped and I can find a way to finish this up.

src/CovidAppointmentTable.js
orzechowskid: interesting, ok!  I'll see if I can find (or derive) a more unique key for each list item.

livgust (approved):

Issue #8, opened 2021-02-12
Link phase & eligibility information


livgust:
Put phase and eligibility information somewhere easily noticeable. Preferably get it straight from the state's website or link to it so that we don't have to keep it updated manually.

jrogers97:
@livgust it looks like the state website uses this google form inside an iframe to let the user check their eligibility - https://docs.google.com/forms/d/e/1FAIpQLSeOcfCPu_afvFILB_nXZz3v9VMaFZZQMgGbKGiU7o3VBr5m7Q/viewform?embedded=true. 
Maybe a "Check eligibility" button that opens up a dialog with this form embedded?

harcod:
@livgust I think that this issue can be closed by PR #31

PR #10, opened 2021-02-12
Add eligibility modal

jrogers97:
Issue #8 

Adds a button below the site header that opens a modal allowing the user to check their eligibility.
Uses the same form as https://www.mass.gov/covid-19-vaccine does, embedded in an iframe.

Screen Shot 2021-02-12 at 6 30 49 PM
Screen Shot 2021-02-12 at 6 30 44 PM


src/App.js
```diff
@@ -23,12 +24,25 @@ function App() {
...
+						<h1 className={classes.heading}>
+							MA Covid Vaccine Appointments
+						</h1>
+						<StateEligibility />
 						<CovidAppointmentTable />
-						<Typography variant="caption" display="block" gutterBottom>
-						This site is not affiliated with or endorsed by the Commonwealth of Massachusetts.<br/>
-This site is for informational purposes only.   Not all vaccination locations are tracked and the information may not be complete or accurate.<br/>
-							Copyright &#169; {new Date().getFullYear()} Olivia Adams. All rights reserved.
+						<Typography
```
jrogers97, line 32: sorry, auto-formatter got a little crazy

harcod:
Perhaps this feature should be accessible from the "menu" rather than a button beneath the site title.

livgust (commented):
This looks great, thanks! one question - do you think there's a way to notice when the user is at the last page and either cut off the buttons, reroute them to close the modal, or put some other notice to click outside of the modal to exit? It's a little weird for them to click "Submit" after seeing whether they are or aren't eligible, esp. when it tells them to go to a mysterious "Step 2"

livgust:
> Perhaps this feature should be accessible from the "menu" rather than a button beneath the site title.

We have another PR dealing with filters, so I think we'll actually need to re-design that portion of the site in general...

jrogers97:
@livgust I don't think we can add logic based on which page they're on in the form, because it's all within an iframe. (At least as far as I know). Agreed though that the UX feels a bit off. If that's a dealbreaker maybe we can just link out to 
 https://www.mass.gov/covid-19-vaccine, which is honestly way more informative (says right off the bat what group is currently eligible) and has the same form embedded

livgust:
@jrogers97 Thanks again for doing this. If you could resolve the conflicts I'll sign off and merge.

livgust (approved):

Issue #4, opened 2021-02-12
Allow filtering to type of vaccine


livgust:
Some people are looking to book their second appointment and have to find one for the same vaccine manufacturer as their first appointment.

This would likely mean an update to the back-end scrapers to ensure that they pass that information if available.

If we add enough filters and sorters, we likely will want some kind of gear or menu for filters so they don't clutter the page.

linusmarco:
I'm working on this one. Should have a PR in this weekend. As part of this I'm doing a little refactor to group all filters into a component t separate from the main site list.

PR #13, opened 2021-02-13
Add vaccine type filter, move all filters to separate component

linusmarco:
This PR: 

- Adds a vaccine type filter (Implements #4)
- Moves the added filter and the existing availability filter to their own component/panel
  - This new panel lives on the left side of the screen for md+ size screens (desktop, probably) and at the top on smaller screens (mobile, probably)
  - On larger screens, the filters are sticky so that they follow when the user scrolls down
- Removes the `.eslintcache` file from version control
- Fixes a few React warnings

Larger screen:
large-screen

Smaller screen:
small-screen


harcod:
**Bug Report:** "Amherst Regional High School" gets replicated in the listing when switching back and forth between "Pfizer" and "Moderna" filter.  In addition, the replicas don't have any "More Information" attached to them. Please note that the current data feed has two separate cards for "Amherst Regional High School" even on the live site.

src/components/FilterPanel.js
```diff
@@ -0,0 +1,112 @@
...
+
+    const [hasAppointments, setHasAppointments] = useState(true);
+    const [vaxType, setVaxType] = useState('Any');
+
+    const { data, onChange } = props;
+
+    useEffect(() => {
+        onChange({
+            hasAppointments: hasAppointments ?
+                d => !!d.hasAppointments === hasAppointments
```
jrogers97, line 41: Isn't this function essentially the same as `d => !!d.hasAppointments`? Because it's only used when `hasAppointments` is true, so  `!!d.hasAppointments === hasAppointments` is always just checking that `!!d.hasAppointments` is true.

Also, to avoid the ternary, this whole value could be simplified to `hasAppointments: d => !hasAppointments || !!d.hasAppointments`

src/components/FilterPanel.js
```diff
@@ -0,0 +1,112 @@
...
+                :
+                () => true,
+            vaxType: vaxType !== 'Any' ?
```
jrogers97, line 44: Same idea here, I think the ternary is slightly harder to understand and could be replaced with `vaxType: d => vaxType === 'Any' || (d.extraData && d.extraData['Vaccinations offered'] && d.extraData['Vaccinations offered'].includes(vaxType))`.

I guess it comes down to preference, maybe the one-liner is harder to read, in which case ignore me!

src/components/FilterPanel.js
linusmarco: Good point. This was some real galaxy brain code...

src/components/FilterPanel.js
linusmarco: To me this feels a little less transparent, but if others prefer this way too I could change it.

johnhawkinson: I think pretty much any use of the ternary operator that isn't for short simple expressions should be avoided, and it definitely shouldn't be combined with arrow functions. I would move the test into the function, and further simplify a bit:

```js
d => {
  if (vaxType !== 'Any') {
    let vaxOffered = d.extraData && d.extraData['Vaccinations offered'];
    return vaxOffered && vaxOffered.includes(vaxType);
  } else {
    return true;
  }
```

(untested, but I think that's equivalent)

linusmarco:
> **Bug Report:** "Amherst Regional High School" gets replicated in the listing when switching back and forth between "Pfizer" and "Moderna" filter. In addition, the replicas don't have any "More Information" attached to them. Please note that the current data feed has two separate cards for "Amherst Regional High School" even on the live site.

I think the only real solution to this issue is to resolve this on the back end by giving each site a unique ID. Below are the two sites causing problems. The issue seems to be that they have a morning and an evening clinic listed separately so the key based on the basic site info (name, city) was not unique. For now I generated a key for each site based on the site's whole object. Not ideal for efficiency reasons but it will work for now.

```
        {
            "name": "Amherst Regional High School",
            "street": "21 Matoon St",
            "city": "Amherst",
            "zip": "01002",
            "availability": {
                "02/18/2021": {
                    "hasAvailability": true,
                    "numberAvailableAppointments": 365,
                    "signUpLink": "https://www.maimmunizations.org/client/registration?clinic_id=1070"
                }
            },
            "hasAvailability": true,
            "extraData": {
                "Vaccinations offered": "Pfizer-BioNTech COVID-19 Vaccine",
                "Age groups served": "Adults, Seniors",
                "Services offered": "Vaccination",
                "Clinic Hours": "09:00 am - 11:45 am"
            }
        },
```

and 

```
        {
            "name": "Amherst Regional High School",
            "street": "21 Mattoon Street",
            "city": "Amherst",
            "zip": "01002",
            "availability": {
                "02/17/2021": {
                    "hasAvailability": true,
                    "numberAvailableAppointments": 229,
                    "signUpLink": "https://www.maimmunizations.org/client/registration?clinic_id=1055"
                }
            },
            "hasAvailability": true,
            "extraData": {
                "Vaccinations offered": "Pfizer-BioNTech COVID-19 Vaccine",
                "Age groups served": "Adults, Seniors",
                "Services offered": "Vaccination",
                "Clinic Hours": "01:00 pm - 02:45 pm"
            }
        },
```

src/components/FilterPanel.js
linusmarco: fixed

src/components/FilterPanel.js
linusmarco: Just as you have an aversion to ternary operators, I have one for assignment expressions with `&&` or `||`. They always require a second look. As a compromise, how about we go fully explicit?

```js
           vaxType: d => {
                if (vaxType === 'Any') {
                    return true;
                } else {
                    if (d.extraData && d.extraData['Vaccinations offered']) {
                        return d.extraData['Vaccinations offered'].includes(vaxType);
                    } else {
                        return false;
                    }
                }
            }
```

johnhawkinson: Err, to be clear, the concern is not the ternary operator — it's the ternary expression being split over 9 lines that include an arrow function, where the test is not obviously better outside the arrow function than in.

Why is `&&` ok in
```js
if (d.extraData && d.extraData['Vaccinations offered']) {
```
but not in
```js
return vaxOffered && vaxOffered.includes(vaxType);
```
?

Also, part of the point of the simplification (and introduction of `let vaxOffered`) was to avoid searching the `d.extraData` dictionary twice (and cut down line length and cognitive load). Yours undoes that.

If you really like `?:` and dislike `&&` I'd replace
```js
    return vaxOffered && vaxOffered.includes(vaxType);
```
with
```js
    return vaxOffered ? vaxOffered.includes(vaxType) : false;
```
which, although silly to my eye, would seem better than having an entire 5-line if/else block to accomplish the same thing.

But I am In No Way King, and your revised solution eliminates the mass confusion point of the long ternary-operator-with-arrow-function-within, so is definitely an improvement over the original.

linusmarco: I'm not trying to pick a fight here, man--just addressing readability. I'll go with the most recent revision with the explicit if-else's so we're not running in circles.

harcod:
> **Bug Report:** "Amherst Regional High School" gets replicated in the listing when switching back and forth between "Pfizer" and "Moderna" filter. In addition, the replicas don't have any "More Information" attached to them. Please note that the current data feed has two separate cards for "Amherst Regional High School" even on the live site.

### Clarification and Steps to Reproduce Bug:  

1. Load the page the first time.  The vax type "Any" is selected by default.
There are two "Amherst Regional High School" entries in the list. This is correct because they are actually different times.

2. Select filter Vax Type = "Pfizer".
Now there are **three** "Amherst Regional High School" entries in the list. 

3. Select filter Vax Type = "Moderna".
Now there are **two** "Amherst Regional High School" entries in the list. Neither of these two entries have any "More Information" available.

4. Select filter Vax Type = "Pfizer"
Now there are **four** "Amherst Regional High School" entries in the list. 

If you continue to switch back and forth, more and more entries are created for "Amherst" in the listing.

linusmarco:
> > **Bug Report:** "Amherst Regional High School" gets replicated in the listing when switching back and forth between "Pfizer" and "Moderna" filter. In addition, the replicas don't have any "More Information" attached to them. Please note that the current data feed has two separate cards for "Amherst Regional High School" even on the live site.
> 
> ### Clarification and Steps to Reproduce Bug:
> 1. Load the page the first time.  The vax type "Any" is selected by default.
>    There are two "Amherst Regional High School" entries in the list. This is correct because they are actually different times.
> 2. Select filter Vax Type = "Pfizer".
>    Now there are **three** "Amherst Regional High School" entries in the list.
> 3. Select filter Vax Type = "Moderna".
>    Now there are **two** "Amherst Regional High School" entries in the list. Neither of these two entries have any "More Information" available.
> 4. Select filter Vax Type = "Pfizer"
>    Now there are **four** "Amherst Regional High School" entries in the list.
> 
> If you continue to switch back and forth, more and more entries are created for "Amherst" in the listing.

@harcod are you still seeing this after the most recent commits to address the issue? This was caused by dupe keys. See my previous comment on how I fixed it by creating a unique key. I don't see this bug after 158e646, but please let me know if you still do

harcod:
@linusmarco Fixed now. ( Sorry.  For some reason, I didn't have that one update the fixed dupe keys. )

src/components/FilterPanel.js
```diff
@@ -0,0 +1,111 @@
...
+                        return d.extraData['Vaccinations offered'].includes(vaxType);
+                    } else {
+                        return false;
+                    }
+                }
+            }
+        })
+    }, [onChange, hasAppointments, vaxType])
+
+    const vaxTypes = Array.from(
```
livgust, line 55: I'm a little nervous here. Only the MA vaccination sites will have that "Vaccinations offered" information that we scrape up; no other sites will have it. Therefore by default we'll show them in the "Any" bucket which is misleading. It might be worth entering manually what sites (outside of the state sites) have which vaccines, but then that's something we would have to manually update if it ever changes.

Curious if you (or anyone else) have thoughts.

livgust:
OK so I mentioned this morning that I'm a little nervous about the design, since we have proximity filtering coming through soon too as well. What do you think about this? It's a common design pattern on other websites with filtering.
(Also ignore my bad Paint skills, replace [] with checkboxes, () with radio buttons, and use Material UI text styling)

Desktop:
image

Mobile:
image

livgust:
I like [this component](https://material-ui.com/components/drawers/#responsive-drawer) for a side bar - it's basically the same as what you did but it looks a little more "separate" which I like. Thoughts?

livgust (changes requested):

linusmarco:
This has now been updated to use the new responsive drawer for the filters and now uses the checkbox-based design that @livgust proposed:

Desktop:
image

Mobile:
(filter panel closed)
image
(filter panel open)
image

livgust:
LOVE THIS so far!!!

I think next steps are:
1) confirming design with Daniel Cho (danieljaecho93)
2) adding proximity filtering (zip code search), which we already have some code for in another PR

livgust:
In true Olivia fashion, I was unable to pick this up today. Womp womp.

linusmarco:
@danieljaecho93, thoughts on the design here?

danieljae:
@linusmarco Looks great :) Something to consider on my end is what sorts of different categories do we want to display for the filters-for instance, I'm wondering how much people would want to filter for places with no available appointments? But I will do some research on my own and look into that, this will definitely work for the time being!

Issue #3, opened 2021-02-12
Filtering by location


livgust:
I don't have a particular solution in mind, but I know a standard one would be to have a ZIP filter and then have a radius of X miles.

dldereklee:
Can add the Distance Matrix API from Google, but would need to have geo data (lat, long) for each location from the scrapper

linusmarco:
The distance matrix API does allow you to pass addresses. Although it would be more reliable if the back end returned lat/lon for each location. 

https://developers.google.com/maps/documentation/distance-matrix/overview

johnhawkinson:
> Can add the Distance Matrix API from Google, but would need to have geo data (lat, long) for each location from the scrapper

We need it for other purposes (Visualization, see https://github.com/livgust/covid-vaccine-scrapers/issues/32).
My intention was to cache it in a separately maintained `geo.json` that clients would download, and just periodically geocode new site addresses, because new sites are rare.

But maybe there's a better architecture.

linusmarco:
Until we get the vaccination site geocoding implemented on the back end, we could add a simple zipcode (or maybe county, which we can back out from zipcode) filter.

dldereklee:
If you have the street and the zip of the location, the geocoding service is pretty stable. Can add it to the scrapper which currently gets street and zip. 
```
const { Client } = require("@googlemaps/google-maps-services-js");

const getGeocode = async (street, zip) => {
  const address = `${street},${zip}`;
  const client = new Client();
  try {
    const resp = await client.geocode({
      params: {
        address,
        key: process.env.GOOGLE_API_KEY,
      },
    });
    return resp.data;
  } catch (e) {
    console.error(e.response.data);
  }
};
```
Response
```
{
    "results": [
        {
            "address_components": [REDACTED],
            "formatted_address": "153 Chestnut St, Southbridge, MA 01550, USA",
            "geometry": {
                "bounds": {
                    "northeast": {
                        "lat": 42.0687815,
                        "lng": -72.0273536
                    },
                    "southwest": {
                        "lat": 42.0682489,
                        "lng": -72.02792219999999
                    }
                },
                "location": {
                    "lat": 42.0685246,
                    "lng": -72.0276497
                },
                "location_type": "ROOFTOP",
                "viewport": {
                    "northeast": {
                        "lat": 42.0698641802915,
                        "lng": -72.02628891970849
                    },
                    "southwest": {
                        "lat": 42.0671662197085,
[...]

PR #15, opened 2021-02-14
Ranking by Location Distance

dldereklee:
I gave it a try to implement ranking by location distance (#3). This PR allows users to:

1. Input a zipcode and get ranked locations based on distance
2. Use browser `navigator.geolocation`  to get user location and get ranked locations based on distance

In order for this to work, the backend needs to return entries with a `latitude`  and `longitude` property (so I wouldn't merge this unless the backend changes are made). The frontend calculates the distance using the `geolib` library which takes two lat/long coordinates and calculates the distance between. If there is no associated lat/long with the location, the location is filtered out from the list. 

I used https://public.opendatasoft.com/api to convert the zipcodes to lat/long. It's the only API I could find that did not require API keys. The dataset is here (https://public.opendatasoft.com/explore/dataset/us-zip-code-latitude-and-longitude/api/)

image

Let me know if you have any feedback. Also forgive the formatting. I didn't add Prettier in yet to avoid having entire files changed solely due to formatting.


src/CovidAppointmentTable.js
```diff
@@ -65,22 +87,51 @@ export default function CovidAppointmentTable() {
...
 	const formattedData = sortAndFilterData(
 		transformData(data),
 		sortInfo,
 		onlyShowAvailable
 	);
 
+	const getCurrentLocation = () => {
+		setZipcode("")
+		navigator.geolocation.getCurrentPosition((position) => {
+		  setCurrentLocation({latitude: position.coords.latitude, longitude: position.coords.longitude});
```
livgust, line 119: are there other keys besides latitude and longitude? If not, you could just do
```
setCurrentLocation(position.coords);
```

src/CovidAppointmentTable.js
```diff
@@ -65,22 +87,51 @@ export default function CovidAppointmentTable() {
 	const [onlyShowAvailable, setOnlyShowAvailable] = useState(true);
 
 	useEffect(() => {
+		resetData()
+	}, []);
+
+	useEffect(()=> {
+		setLoading(true)
+		setData(data => data.filter(entry => entry.latitude && entry.longitude)
```
livgust, line 95: Filtering out any location without a latitude & longitude seems heavy-handed to me.

src/CovidAppointmentTable.js
```diff
@@ -65,22 +87,51 @@ export default function CovidAppointmentTable() {
...
+			.map(entry=> ({...entry, distance: getDistance(currentLocation, {latitude:entry.latitude || 0, longitude: entry.longitude || 0}) * 0.000621371 }))
```
livgust, line 96: I don't think that `data` should be modified here; it should represent what comes from the API endpoint. The more appropriate place to do sorting and filtering computations is within the `sortAndFilterData` function that is called below.

livgust: Also, we shouldn't sort by anything or show any distance if the location is not set.

src/CovidAppointmentTable.js
```diff
@@ -90,13 +141,24 @@ export default function CovidAppointmentTable() {
...
 			/>
+			<div >
+				<TextField id="outlined-basic" label="Enter Zipcode" variant="outlined" onChange={e => setZipcode(e.target.value)} value={zipcode}/>
+				<Button className={classes.filterButtons} variant="contained" color="secondary" onClick={getZipcodeLocation}>Search By Zip</Button>
+				<Button className={classes.filterButtons} variant="contained" color="secondary" startIcon={<GpsFixedIcon/>} onClick={getCurrentLocation}>Use My Location</Button>
+			</div>
+			{!loading && data.length === 0 ?
+				<div className={classes.notFoundBox}>
+					<h2 className={classes.locationsNotFound}>No Locations Found</h2>
+					<Button className={classes.filterButtons} variant="contained" color="secondary" onClick={resetData}>Show All Locations</Button>
```
livgust, line 152: Since this will be a third type of filter (availability, vaccine type, and now ZIP), I've suggested an alternative, more unified UI. LMK what you think: #13

src/CovidAppointmentTable.js
```diff
@@ -90,13 +141,24 @@ export default function CovidAppointmentTable() {
...
+				</div> :(
+			<Loader loaded={!!data && data.length > 0 && !loading}>
 			{formattedData.map((entry) => {
 				return (
 					<div className={classes.cardBox}>
 						<Card>
 							<CardHeader
 								title={<div>{entry.location}</div>}
-								subheader={<div>{entry.city}</div>}
+								subheader={<><div>{entry.city}</div>{entry.distance && <div><RoomOutlined/>{Number.parseFloat(entry.distance).toFixed(1)} miles away</div>}</>}
```
livgust, line 161: I made a comment above, but I don't think we should show this unless a ZIP is set.

livgust (changes requested):
Hi Derek! Thanks so much for this, I really like what you've done. Please review my comments and let me know what you think. They're mainly around:

- separating data fetching from filtering and sorting
- not showing distance data unless a ZIP is entered
- choosing a UI consistent with other filters on the page

In addition, could you install `prettier` and run that on your edited files for styling consistency? Thank you!!

src/CovidAppointmentTable.js
dldereklee: This is one that I was debating. What should we show for locations with no lat/long? I would include them in the list, but I don't know if that's more confusing to the user.

johnhawkinson: Silently removing things is more confusing than returning things adequately labeled.
Putting them last sounds good to me — do we have some expectation that this would be a frequent problem? Last I had looked everything came with an address.

dldereklee: I put it there to catch the cases where the entry is missing the `latitude` or `longitude` coordinate since you cannot compute the distance from the user when there is a missing coordinate.

livgust: Instead of filtering it out, we should just not compute the distance when lat/long is missing. And then if there is no `distance` key, don't show it in the UI.
That also would help with the case where you haven't entered a ZIP - no locations have a `distance` key and so the UI element doesn't appear.

src/CovidAppointmentTable.js
dldereklee: There are additional keys such as `accuracy`

src/CovidAppointmentTable.js
dldereklee: What if they use the `Use My Location`? That would provide the coordinates without the zipcode input being provided.

dldereklee:
Related PR: 
https://github.com/livgust/covid-vaccine-scrapers/pull/72

livgust:
FYI @harcod is working on merging this work in with the updated UX design (side bar)

dldereklee:
@livgust Should I close this since it was implemented in another PR?

harcod:
It appears that this functionality was independently developed and recently merged via PR #115.
I'm going to close this issue, but feel free to reopen it if you disagree.

PR #17, opened 2021-02-15
Add prettier config and update docs

MattRC7:
- [X] Add `.prettierrc.json` and `.prettierignore` files to project root.
- [X] Run `npx prettier --write *` to enforce prettier formatting for all files.
- [X] Update README with instructions for building/running the app.


MattRC7:
Do you want to add [husky](https://typicode.github.io/husky/) to set up a pre-commit hook that will either run `prettier --write` to ensure code is formatted, or at least `prettier --check` to validate that the code is formatted properly?

carloshgaray:
> Do you want to add husky to set up a pre-commit hook that will either run prettier --write to ensure code is formatted, or at least prettier --check to validate that the code is formatted properly?


Yes, can we also add airbnb prettier and eslint styles?

livgust (approved):
Thank you! And yes, I'd love to automate running prettier/eslint/etc on every PR

PR #20, opened 2021-02-15
Markdown parsed, need to render correctly

tom-ricci:
I've parsed the markdown sites use in the additional information section, but I'm stuck on how to render the parsed information as actual html instead of text. If I should be using something other than a PR since it's not complete let me know.


harcod:
To display as actual HTML, you can use `dangerouslySetInnerHTML`.  As the name states, this is dangerous.   Either you or the scrapers (but not both) should be encoding and HTML entities in the data retrieve from servers.

Replace lines 116:118 with...

``<div className={classes.extraData} dangerouslySetInnerHTML={{__html: `<b>${key}:</b> ${data[key]}`}}/>``

MattRC7:
Hey, i actually was working on a way to parse markdown into React JSX recently, if I have time tonight i can send you my code snippets

If my solution turns out to be too heavyweight then dangerously setting the HTML ought to work, as long as we do due diligence.

tom-ricci:
> Hey, i actually was working on a way to parse markdown into React JSX recently, if I have time tonight i can send you my code snippets
> 
> If my solution turns out to be too heavyweight then dangerously setting the HTML ought to work, as long as we do due diligence.

Considering the project already uses some dependencies and loads quickly, it probably won't add too much weight. I'll try out the snippets first, thanks!

src/components/MoreInformation.js
```diff
@@ -5,60 +5,119 @@ import ExpandMoreIcon from "@material-ui/icons/ExpandMore";
...
+            newDataEm = newDataEm + "</em>" + splitDataEm[i];
+            isEm = false;
+          } else {
+            newDataEm = newDataEm + "<em>" + splitDataEm[i];
+            isEm = true;
+          }
+          i++;
+        }
+        entry.extraData["Additional Information"] = newDataEm;
+      }
```
MattRC7, line 84: I'd like to recommend you take this and move it into a function that takes a markdown string and parses it, returning the output. Then you can pass that result into the `<ExtraData />` element below. (Pick a better name for the function than my example :) )

```jsx
const parsedExtraData = myFunctionThatParsesMarkdown(entry.extraData)

// ...

<ExtraData data={parsedExtraData} />
```

MattRC7 (commented):
Grabbing my snippets to attach in a moment. Had one suggestion for code organization.

MattRC7:
Here's a snippet of code I've been messing around with for converting Markdown to JSX. Not sure if this is too heavyweight. @livgust idk how concerned you are about dependencies, but `commonmark` is a pretty performant and lightweight markdown parser.

Even if we don't use a library to parse the data, since we're in React world, you could look into doing something similar to the switch statement below to create a JSX tree instead of building a string, then pass that to the `<ExtraData />` to be rendered. That way you don't have to dangerously set the HTML.

```jsx
import React from 'react';
import { Parser, Node } from 'commonmark';

const parser = new Parser();

/** Recursive function that takes nodes parsed by commonmark and converts them to JSX */
const nodeToJSX = (node, key = '') => {
  let nextNode = node.firstChild;
  if (!nextNode) {
    // base case: return the text of nodes without any children.
    return node.literal;
  }

  // iterate through the node's children and convert them to JSX.
  let content = [];
  let uniqueKey = 0; // Arrays of JSX elements need unique keys.
  while (nextNode) {
    content.push(nodeToJSX(nextNode, `${key}-${uniqueKey}`));
    nextNode = nextNode.next;
    uniqueKey += 1;
  }

  // wrap the JSX from the children in an element based on the type of Markdown node.
  // You only need to implement the types you care about such as `<em>`.
  switch (node.type) {
    case 'strong':
      return <strong key={key}>{content}</strong>;
    case 'emph':
      return <em key={key}>{content}</em>;
    case 'heading':
      switch (node.level) {
        case 1:
          return <h1 key={key}>{content}</h1>;
        case 2:
          return <h2 key={key}>{content}</h2>;
        // continued for h3, h4, etc...
        default:
          return <h1 key={key}>{content}</h1>;
      }
    // See docs for other elements: https://github.com/commonmark/commonmark.js#usage 
    default:
      return <div key={key}>{content}</div>;
  }
};

export const MyComponent = (markdownString) => {
  return nodeToJSX(parser.parse(markdownString));
};
```

Let me know what you think, and if you have questions!

livgust:
Not concerned about dependencies for the front end!

Curious, how many sites are we trying to do formatting for? Is it always presented the same way? If so, might be worth doing just straight regexes instead of interpreting markdown. Ignore me if I'm misunderstanding or just plain wrong lol, I'm looking at this on my phone

src/components/MoreInformation.js
```diff
@@ -5,60 +5,119 @@ import ExpandMoreIcon from "@material-ui/icons/ExpandMore";
...
-                <div className={classes.extraData}>
-                    <b>Address:</b> {entry.streetAddress}, {entry.city}, MA{" "}
-                    {entry.zip}
-                </div>
-                <ExtraData data={entry.extraData} />
-            </AccordionDetails>
-        </Accordion>
-    );
+  const classes = useStyles();
+  if (!(entry.extraData == null)) {
```
livgust, line 30: why not
```
if (entry.extraData)
```
?

If you need to check for null, use `===` instead of `==`

src/components/MoreInformation.js
```diff
@@ -5,60 +5,119 @@ import ExpandMoreIcon from "@material-ui/icons/ExpandMore";
...
+    //reggie lewis uses markdown but not for emphasis so it skips that site
+    if (!(entry.location == "Reggie Lewis State Track Athletic Ctr")) {
```
livgust, line 32:
```
if (entry.location !== "Reggie Lewis ...") {
```

src/components/MoreInformation.js
```diff
@@ -5,60 +5,119 @@ import ExpandMoreIcon from "@material-ui/icons/ExpandMore";
...
+      var currentData = entry.extraData["Additional Information"];
+      if (!(currentData == null)) {
```
livgust, line 34:
```
if (currentData) {
```

src/components/MoreInformation.js
```diff
@@ -5,60 +5,119 @@ import ExpandMoreIcon from "@material-ui/icons/ExpandMore";
...
-                </div>
-                <ExtraData data={entry.extraData} />
-            </AccordionDetails>
-        </Accordion>
-    );
+  const classes = useStyles();
+  if (!(entry.extraData == null)) {
+    //reggie lewis uses markdown but not for emphasis so it skips that site
+    if (!(entry.location == "Reggie Lewis State Track Athletic Ctr")) {
+      var currentData = entry.extraData["Additional Information"];
```
livgust, line 33: You should go through and change all of these instances of `var` to either `const` or `let`.
The difference:
`const` is if you never reassign the variable. For example, `currentData` is never reassigned so it can be a `const`.
`let` allows reassignment. Gotcha: if you are setting object properties, the object can be a const. For example,
```
const objectExample = {};
objectExample.key = "value";
```
is legal and appropriate (imo).

src/components/MoreInformation.js
```diff
@@ -5,60 +5,119 @@ import ExpandMoreIcon from "@material-ui/icons/ExpandMore";
...
+      if (!(currentData == null)) {
+        var splitDataEmStr = currentData.split("***");
+        var i = 1;
+        var newDataEmStr = splitDataEmStr[0];
+        var isEmStr = false;
+        while (i < splitDataEmStr.length) {
```
livgust, line 39: I'd change this to a `for` loop since your variable `i` always increments and has a discrete end.

src/components/MoreInformation.js
```diff
@@ -5,60 +5,119 @@ import ExpandMoreIcon from "@material-ui/icons/ExpandMore";
...
+              splitDataEmStr[i];
+            isEmStr = true;
+          }
+          i++;
+        }
+        var splitDataStr = newDataEmStr.split("**");
+        var i = 1;
+        var newDataStr = splitDataStr[0];
+        var isStr = false;
+        while (i < splitDataStr.length) {
```
livgust, line 56: Same comment as in line 39

src/components/MoreInformation.js
```diff
@@ -5,60 +5,119 @@ import ExpandMoreIcon from "@material-ui/icons/ExpandMore";
...
+              splitDataStr[i];
+            isStr = true;
+          }
+          i++;
+        }
+        var splitDataEm = newDataStr.split("*");
+        var i = 1;
+        var newDataEm = splitDataEm[0];
+        var isEm = false;
+        while (i < splitDataEm.length) {
```
livgust, line 73: Same comment as in line 39

src/components/MoreInformation.js
livgust: In addition, this function could benefit from lots of comments! Don't be shy!

livgust (changes requested):
Hey @tom-ricci, lots of nitpicky comments here. A lot of my asks are Javascript-specific kinds of comments; I'm assuming you're not super familiar with the language but I apologize if not.

Thanks for doing this!

tom-ricci:
> Hey @tom-ricci, lots of nitpicky comments here. A lot of my asks are Javascript-specific kinds of comments; I'm assuming you're not super familiar with the language but I apologize if not.
> 
> Thanks for doing this!

Ok, I'll take a look at them. While I do have javascript experience I'd say it's definitely one of the ones I'm less familiar with.

tom-ricci:
Ok, I've made the changes from the comments and fixed the problem with rendering. I went with escaping <> and setting the inner html dangerously rather than the code snippet from @MattRC7 (it seemed a bit quicker and easier to implement)

Also allowed reggie to be parsed now, I accidently parsed them and noticed they were using markdown like the rest of the sites and I just read it wrong originally. I'm guessing you guys just need to review the newer commits now and either merge or give other comments but if not lmk

src/components/MoreInformation.js
```diff
@@ -53,12 +55,100 @@ function ExtraData({ data }) {
...
+    if (indexOfLAB !== -1) {
+        workingData1 = currentData.replaceAll("<", "&#60;");
+    } else {
+        workingData1 = currentData;
+    }
+    if (indexOfRAB !== -1) {
+        workingData2 = workingData1.replaceAll(">", "&#62;");
+    } else {
+        workingData2 = workingData1;
+    }
```
MattRC7, line 100: This is a good start but think we're still vulnerable to a Cross-Site Scripting attack here. Here's the [OWASP Cheat Sheet](https://cheatsheetseries.owasp.org/cheatsheets/Cross_Site_Scripting_Prevention_Cheat_Sheet.html#rule-1-html-encode-before-inserting-untrusted-data-into-html-element-content) for preventing XSS attacks. Rule 1 pertains to what we're doing, inserting into plain HTML elements. There are a few characters we need to look out for, and the cheat sheet recommends using a reliable library if possible.

If we can use a library, [DOMPurify](https://github.com/cure53/DOMPurify) seems like a well maintained, lightweight option. If we want to roll our own, let's make sure we cover at least the characters suggested on the cheat sheet: `&, <, >, ", '`.

@livgust do you have a preference here?

src/components/MoreInformation.js
```diff
@@ -53,12 +55,100 @@ function ExtraData({ data }) {
...
+        if (isEmStr) {
+            newDataEmStr = newDataEmStr + "</span></em>" + splitDataEmStr[i];
+            isEmStr = false;
+        } else {
+            newDataEmStr =
+                newDataEmStr +
+                '<em><span style="color: #F50057">' +
+                splitDataEmStr[i];
+            isEmStr = true;
+        }
```
MattRC7, line 118: Optional: you could simplify a little by using template strings + wrapping the em in one pass:

```suggestion
        if (isEmStr) {
            newDataEmStr = newDataEmStr + splitDataEmStr[i];
            isEmStr = false;
        } else {
            newDataEmStr = newDataEmStr + `<em><span style="color: #F50057">${splitDataEmStr[i]}</span></em>`;
            isEmStr = true;
        }
```

src/components/MoreInformation.js
```diff
@@ -53,12 +55,100 @@ function ExtraData({ data }) {
...
+        if (isStr) {
+            newDataStr = newDataStr + "</span>" + splitDataStr[i];
+            isStr = false;
+        } else {
+            newDataStr =
+                newDataStr +
+                '<span class="formattedExtraData">' +
+                splitDataStr[i];
+            isStr = true;
+        }
```
MattRC7, line 136: Same thing, if you want to:

```suggestion
        if (isStr) {
            newDataStr = newDataStr + splitDataStr[i];
            isStr = false;
        } else {
            newDataStr = newDataStr + `<span class="formattedExtraData">${splitDataStr[i]}</span>`;
            isStr = true;
        }
```

Also, do you know why we're using a span with a class for bold instead of the `<strong></strong>` HTML element?

src/components/MoreInformation.js
```diff
@@ -53,12 +55,100 @@ function ExtraData({ data }) {
...
+    //iterates through the split data
+    //same formatting rules as 85
+    for (i = 1; i < splitDataEm.length; i++) {
+        if (isEm) {
+            newDataEm = newDataEm + "</em>" + splitDataEm[i];
+            isEm = false;
+        } else {
+            newDataEm = newDataEm + "<em>" + splitDataEm[i];
+            isEm = true;
+        }
```
MattRC7, line 151:
```suggestion
        if (isEm) {
            newDataEm = newDataEm + splitDataEm[i];
            isEm = false;
        } else {
            newDataEm = newDataEm + `<em>${splitDataEm[i]}</em>`;
            isEm = true;
        }
```

MattRC7 (commented):
Left a few more comments, let me know what you think!

src/components/MoreInformation.js
```diff
@@ -24,6 +24,8 @@ const useStyles = makeStyles((theme) => ({
 
 export default function MoreInformation({ entry }) {
     const classes = useStyles();
+    if (entry.extraData) {
+    }
```
livgust, line 28: is this intended?

src/components/MoreInformation.js
```diff
@@ -53,12 +55,100 @@ function ExtraData({ data }) {
...
+            //formats additional information instead of just adding it to the more information section
+            if (key === "Additional Information") {
+                const currentData = data[key];
+                if (currentData) {
+                    var finalData = parseMD(currentData);
+                    //adds the parsed data as straight html
+                    elements.push(
+                        <div
+                            className={classes.extraData}
+                            dangerouslySetInnerHTML={{
```
livgust, line 67: Sorry, I might've missed when we decided to use dangerouslySetInnerHTML but I think we should avoid this at all costs. As @MattRC7 mentioned, this is a security risk.

src/components/MoreInformation.js
```diff
@@ -53,12 +55,100 @@ function ExtraData({ data }) {
...
         return elements;
     }
 }
+
+function parseMD(currentData) {
+    //encodes angle brackets to escape data
+    let workingData1;
+    let workingData2;
+    let indexOfLAB = currentData.indexOf("<");
+    let indexOfRAB = currentData.indexOf(">");
```
livgust, line 90: these don't have values reassigned so they should be `const`s

src/components/MoreInformation.js
livgust, line 100: seems we can call `replaceAll` on a string without any occurrences of the replacement, so you can just always call that function regardless of if there's a > or <.

I agree with @MattRC7; if we can just avoid using `dangerouslySetInnerHTML` then that solves our problem. I'm cool with using a third-party library if there's one that looks like a good fit; makes our code less bulky.

src/components/MoreInformation.js
livgust: To add to this, using += would make it even more concise while still retaining clarity.

src/components/MoreInformation.js
```diff
@@ -53,12 +55,100 @@ function ExtraData({ data }) {
...
+        } else {
+            newDataEmStr =
+                newDataEmStr +
+                '<em><span style="color: #F50057">' +
+                splitDataEmStr[i];
+            isEmStr = true;
+        }
+    }
+    //parses bold next, splitting data at each "**"
+    let splitDataStr = newDataEmStr.split("**");
```
livgust, line 121: `const`

src/components/MoreInformation.js
```diff
@@ -53,12 +55,100 @@ function ExtraData({ data }) {
...
+        } else {
+            newDataStr =
+                newDataStr +
+                '<span class="formattedExtraData">' +
+                splitDataStr[i];
+            isStr = true;
+        }
+    }
+    //parses italics last, splitting data at each "*"
+    let splitDataEm = newDataStr.split("*");
```
livgust, line 139: `const`

livgust (changes requested):

MattRC7:
Hey @tom-ricci , do you mind if i clone your repo and contribute to this branch? I think I have a way to avoid dangerously setting HTML without changing your code too much, but it's gonna be a lot to type out here, easier to show visually in a commit.

tom-ricci:
@MattRC7 Sure, you can commit. Thanks!

MattRC7:
@livgust I pushed the following two commits which sanitize the HTML we are generating before returning it to be dangerously set. This should be sufficient protection against any malicious strings within the markdown: https://github.com/livgust/macovidvaccines.com/pull/20/files/fc8c9113e30353f5f433ccb09f3d2cbd1fd6f9a0..623b2e874a67da25149edf1afd8625529b63f3e9 . If you're content with this, and @tom-ricci  takes care of the rest of your comments, we should be good to go.

If we want to avoid `__dangerouslySetInnerHTML` all-together, we're going to have to parse the Markdown and generate the HTML a different way: either we use a more sophisticated `.split()` using a regex with capture groups, or we pull in a parser like `commonmark`. Either way, we would then need to create the output in a single pass and output a tree of JSX nodes instead of an HTML string. That's going to be a significant re-write compared to what we currently have. Up to you @livgust , we can rewrite, we can go with this, or we can merge this first and one of us can work on a more complex parser in a different PR.

Thanks for bearing with me @tom-ricci , this website is a prime target for malicious agents (publicly advertised, accessed by many users without internet literacy, open-source code without a security auditing team, project under a time crunch) so we want to make sure we're covering all our bases on the security front.

tom-ricci:
Ok, I'll take care of the rest of the comments quickly so if we choose to go with this we can merge it soon. Also looking at @MattRC7's comment about using a span rather than strong, I used the span to change the color rather than bold since the titles were already bolded and I wanted to have a difference between the titles and the content but bolding works if we choose to go with that.

Also @MattRC7 it's somewhat ironic because I was going to thank you for bearing with me since I've noticed working on this site is more of a learning experience rather than just coding comfortably, but you're welcome!

MattRC7:
> Ok, I'll take care of the rest of the comments quickly so if we choose to go with this we can merge it soon. Also looking at @MattRC7's comment about using a span rather than strong, I used the span to change the color rather than bold since the titles were already bolded and I wanted to have a difference between the titles and the content but bolding works if we choose to go with that.
> 
> Also @MattRC7 it's somewhat ironic because I was going to thank you for bearing with me since I've noticed working on this site is more of a learning experience rather than just coding comfortably, but you're welcome!

Ahh, that makes sense about changing the color, I like it. If you want to you still can, just leave it as a `strong` tag so screen readers understand. You can still pass it a class attribute with styles to change the color. Lemme know if that's new to you and I'll post a snippet, or you can look for examples in the app.

And no problem, this is right in my wheelhouse so I don't mind!

tom-ricci:
Ok, unless I missed any comments above that should be everything.

MattRC7 (approved):
This lools good to me!

MattRC7:
Hey @livgust , if you have time, can we get a look at this before it gets stale?

livgust:
Sorry @MattRC7 ! I haven't forgotten, we just had a ridiculously busy day yesterday. I'll get to it this afternoon for sure.

src/components/MoreInformation.js
```diff
@@ -20,6 +21,9 @@ const useStyles = makeStyles((theme) => ({
         display: "block",
         "padding-bottom": theme.spacing(1),
     },
+    formattedExtraData: {
+        color: "#AD003B",
```
livgust, line 25: Can we change this to something from the material UI theme? Maybe theme.palette.error.dark?

livgust (changes requested):
Just tested locally and found a couple things:
image
^ bug

image

^ there are 3 locations with this annoying italic + bold text - I think honestly we can just go without the coloring. It's really bright and it looks like in most cases it's already emphasized another way already.

I left a comment asking to change the color to something in the Material UI palette but I think that may be moot if you agree with my judgement call on the above.

Thank you!!

MattRC7:
Hey @tom-ricci , I made the changes Liv suggested, but it won't let me push to your branch now, I'm getting permission denied. Do you have to allow me to again?

tom-ricci:
I'm not sure. I can't actually push from vsc because of permissions either so I think it might be something with my account or something.

I'll check right now, I have some time

tom-ricci:
@MattRC7 Ok I've invited you as a collaborator, try accepting the invite and see if it will work now.

MattRC7:
@livgust pushed changes you requested!

livgust (approved):

Issue #7, opened 2021-02-12
Add context on availability


livgust:
Provide some kind of UI element that is able to provide the following feedback (this is not official copy or anything):

- Sometimes we show availability but no sign-up link. This is because the state website doesn't give us a sign-up link. Why? Don't know.
- Sometimes we show a sign up link but no specific availability (right now there's a message saying "No date-specific data available." This is because we are able to determine that there are available appointments, but we can't tell you when or how many.

harcod:
> Sometimes we show availability but no sign-up link. This is because the state website doesn't give us a sign-up link. Why? Don't know.

I was on a conference call last week with officials (mostly politicians) about the registration on MAImmunizations.org

Once all of the appointments are filled, the site collects some number of "stand-by" people that can be called up when slots open up, then it removes the Sign-up link.  When people that have confirmed appointments cancel their appointment, their site shows that as an available dose.  They internally use this information to contact someone on the "stand-by list" to fill it.

Kathleen-Francis:
I think it would be helpful to have a separate "About" page with this kind of context, FAQ, etc, instead of just the dialog box. It's been a couple years since I've worked with React, but I think it would be pretty simple to add routing and an additional page like that. I've used [React Router](https://reactrouter.com/web/guides/quick-start) before, but looks like [Reach Router](https://reach.tech/router/) might be simpler?

In addition, when the `No date-specific data available.` message is shown, it could have a `?` icon or something that you could click and get a little dialog with additional context.

Kathleen-Francis:
Discussed in [slack](https://macovidvaccines.slack.com/archives/C01NBP1N1G9/p1613399739146900?thread_ts=1613396392.143800&cid=C01NBP1N1G9) and decided to start with adding the help icon/dialog box.

I'm starting to work on this, but can't assign to myself bc I'm not a collaborator.

PR #21, opened 2021-02-16
Add HelpDialog component to give context on availability

Kathleen-Francis:
https://github.com/livgust/macovidvaccines.com/issues/7

HelpDialog component includes help icon that you click to see dialog and "OK" button in dialog for closing (also closes if you click outside).

Used to provide additional context in two cases:
- No date-specific data available
- No sign-up link available

### DIFF
--- a/src/components/Availability.js
+++ b/src/components/Availability.js
@@ -18,8 +18,8 @@ export default function Availability({ entry }) {
             return (
                 <div>
                     No date-specific data available.
-                    <HelpDialog 
-                        title='No date-specific data available.' 
+                    <HelpDialog
+                        title="No date-specific data available."
                         text="We were able to determine that there are available appointments,
                         but we can't tell when or how many. Click the sign up button to learn more
                         from the location's website."
--- a/src/components/Availability.js
+++ b/src/components/Availability.js
@@ -1,3 +1,5 @@
+import HelpDialog from "./HelpDialog";
+
 export default function Availability({ entry }) {
     console.log(entry.hasAppointments);
     if (!entry.hasAppointments) {
@@ -13,7 +15,17 @@ export default function Availability({ entry }) {
             }
         }
         if (!availableSlots.length && entry.hasAppointments) {
-            return <div>No date-specific data available.</div>;
+            return (
+                <div>
+                    No date-specific data available.
+                    <HelpDialog 
+                        title='No date-specific data available.' 
+                        text="We were able to determine that there are available appointments,
+                        but we can't tell when or how many. Click the sign up button to learn more
+                        from the location's website."
+                    />
+                </div>
+            );
         } else {
             return (
                 <div>
--- a/src/components/HelpDialog.js
+++ b/src/components/HelpDialog.js
@@ -1,4 +1,5 @@
 import React from "react";
+import Tooltip from '@material-ui/core/Tooltip';
 import HelpOutlineIcon from '@material-ui/icons/HelpOutline';
 import Dialog from "@material-ui/core/Dialog";
 import DialogTitle from "@material-ui/core/DialogTitle";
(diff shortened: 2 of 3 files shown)


Kathleen-Francis:
Here's what this looks like:
image
image

Feedback welcome on all of it, obviously, but especially on the help info itself.

harcod:
https://www.barnstablecountyhealth.org/uncategorized/does-barnstable-county-keep-a-waitlist-for-people-seeking-covid-19-appointments

The count goes up when someone cancels, then they are drawing from a waitlist.  See the above link for the explanation from Barnstable County.

Kathleen-Francis:
Do you know if that's the case for other locations as well?

harcod:
I'm not sure. But, I would guess that since they are all being processed by PrepMod (maimmunizations.org) that it probably just how it works. I don't think the state has a lot of control over them as they are running sites for many other places. (RI, MD, CA just to name a few)

Perhaps we should just not show availability at locations from PrepMod that have no sign up link.  What is the point really? It's not helping anyone get an appointment.

src/components/Availability.js
```diff
@@ -13,7 +15,17 @@ export default function Availability({ entry }) {
             }
         }
         if (!availableSlots.length && entry.hasAppointments) {
-            return <div>No date-specific data available.</div>;
+            return (
+                <div>
+                    No date-specific data available.
+                    <HelpDialog 
+                        title='No date-specific data available.'
```
livgust, line 22: I like the copy for this one! Could you run `prettier` on this file so that the formatting is consistent (things like ' vs ", for eaxmple)?

src/components/HelpDialog.js
```diff
@@ -0,0 +1,27 @@
...
+import DialogContent from "@material-ui/core/DialogContent";
+import DialogContentText from "@material-ui/core/DialogContentText";
+import Button from "@material-ui/core/Button";
+
+export default function HelpDialog(props) {
+    const [helpOpen, setHelpOpen] = React.useState(false);
+
+    return (<span>
+        <Tooltip title='Click for more info'>
+            <HelpOutlineIcon fontSize='small' onClick={() => setHelpOpen(true)}/>
```
livgust, line 15: Is it possible to make this gray instead of black? It's a little more prominent than necessary imo. I'm not great at design though so feel free to fight me on it ;)

src/components/SignUpLink.js
```diff
@@ -39,7 +40,14 @@ export default function SignUpLink({ entry }) {
                 }
             }
             if (!dateLinkPairs.length) {
-                return "No sign-up link available.";
+                return (
+                    <div>
+                        No sign-up link available.
+                        <HelpDialog 
+                            title='No sign-up link available.' 
+                            text="The state website (maimmunizations.org/) is not currently providing a sign-up link for this location."
```
livgust, line 48: How about:
"We were unable to retrieve a sign-up link for this location. It's likely that this location is pulling from a wait list and is not scheduling any new appointments."

livgust (changes requested):

src/components/HelpDialog.js
Kathleen-Francis: using `color="action"` makes it this dark gray:
image

`color="disabled"` is lighter:
image

can also do pink or purple (bc they're the primary/secondary colors), but I don't think we want that.

livgust: I like it! Even though it's not technically disabled haha.

Kathleen-Francis: currently I have it as `color="action"` (so the darker gray) - https://github.com/livgust/macovidvaccines.com/pull/21/files#diff-58bfc00b595a55aa3aef8861c670d301b372d97bfba315dd11556a247f145bbfR18

should I change it to disabled for the lighter color?

livgust (approved):
Looks great, thanks!
I wonder if there's accessibility work to be done here (adding aria-labels and whatnot) but I'm going to merge this in anyway and put adding a11y eslint rules on my to-do list.

PR #23, opened 2021-02-16
Open feedback link in a new tab

jrogers97:
Tiny change to open feedback mail link in a new tab


src/components/Menu.js
```diff
@@ -98,6 +98,7 @@ export default function ButtonAppBar() {
                     <a
                         href="mailto:macovidvaccines@gmail.com"
                         className={classes.unstyledLink}
+                        target="_blank"
```
livgust, line 101: we should also add 
```
rel="noreferrer"
```

livgust (changes requested):

PR #24, opened 2021-02-16
chore: add nan to package-lock.json

tkjones117:
Issue: I was getting the error `Error: Cannot find module 'nan'` when running `npm ci`. My hypothesis is that I (on a Mac) am seeing this issue and others (PC / Ubuntu users) are not is that the error is thrown from `fsevents` which is platform-specific. 

Ran `npm install` to add `nan` and other related packages to package-lock.json. 
Now I don't get the error when I run `npm ci`.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -10452,8 +10452,7 @@
         "nan": {
             "version": "2.14.2",
             "resolved": "https://registry.npmjs.org/nan/-/nan-2.14.2.tgz",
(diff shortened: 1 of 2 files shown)


livgust (changes requested):
We shouldn't directly modify package-lock.json; it should be a reflection of package.json. If we need that `nan` package, you should run `npm install nan` (or `npm install --save-dev nan` if it's a development dependency and not needed for production) and then check in the changes to both package.json and package-lock.json.

tkjones117:
Got it! Working on fixing...

tkjones117:
I'm now no longer convinced this is a real issue... While I could reliably repro this yesterday, I can't this morning.
I think let's hold off on this until/unless someone else reports seeing this issue, or I run into it again.

If I do, we should push the results of running `npm install --save-dev nan`.

PR #37, opened 2021-02-17
Add warning when sites are restricted

livgust:
If we see text suggesting a site is only open to residents of a certain area, we will show an icon with hover text:
image

When clicked, it will show a dialog with more context.
image

Also slightly refactors the HelpDialog component; make sure that still works (click a ? icon and make sure it shows a Dialog)

Future-proofing: if `data` passes in a "restrictions" item, will use that to as source-of-truth instead of doing a regex against More Information (state-sponsored sites only).

ALSO I changed the default sort to alphabetical based on user feeback.

### DIFF
--- a/src/CovidAppointmentTable.js
+++ b/src/CovidAppointmentTable.js
@@ -11,6 +11,7 @@ import SignUpLink from "./components/SignUpLink";
 import MoreInformation from "./components/MoreInformation";
 import ErrorOutlineIcon from "@material-ui/icons/ErrorOutline";
 import HelpDialog from "./components/HelpDialog";
+import Typography from "@material-ui/core/Typography";
 
 export function transformData(data) {
     return data.map((entry, index) => {
@@ -53,14 +54,15 @@ const useStyles = makeStyles((theme) => ({
         "padding-top": theme.spacing(2),
         "padding-bottom": theme.spacing(2),
     },
-    locationTitle: {
+    restrictionNotice: {
         display: "flex",
         alignItems: "center",
         flexWrap: "wrap",
+        color: theme.palette.text.primary,
     },
     restrictionIcon: {
         color: theme.palette.warning.dark,
-        "padding-left": theme.spacing(1),
+        "padding-right": theme.spacing(1),
     },
 }));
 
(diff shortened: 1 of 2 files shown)


johnhawkinson:
I would suggest we are much better off with clear text adjacent to the site name that says 'MAY BE RESTRICTED' rather than the icon that requires hovering, which sucks (also mobile). The icon could be there too.

macovid mock

livgust:
image
^ modified look

src/CovidAppointmentTable.js
```diff
@@ -60,8 +73,8 @@ export default function CovidAppointmentTable() {
     const [ready, setReady] = useState(false);
     const [errorMessage, setErrorMessage] = useState();
     const [sortInfo, setSortInfo] = useState({
-        sortKey: "hasAppointments",
-        sortAsc: false,
+        sortKey: "location",
```
tkjones117, line 76: Was sorting by location an intended change here?

src/CovidAppointmentTable.js
```diff
@@ -121,29 +134,87 @@ export default function CovidAppointmentTable() {
...
+    if (entry.restrictions) {
+        hasRestriction = true;
+        restrictionText = entry.restrictions;
+    } else if (entry.extraData && entry.extraData["Additional Information"]) {
+        const text = entry.extraData["Additional Information"];
+        if (
+            //"resident" " live" " work" "eligible populations in"
+            text
+                .toLowerCase()
+                .match(/(resident|\slive|\swork|eligible\spopulations\sin)/)
```
tkjones117, line 163: I suspect this will resulting in us flagging more than we need (e.g. `Berkshire Community College Field House` has additional info "FOR THOSE 65 AND ABOVE, OR WHO HAVE TWO OR MORE MEDICAL CONDITIONS ELIGIBLE UNDER PHASE 2. ID will be required." This will be flagged as "may be restricted" even though it is no more restrictive than any other place.

tkjones117: actually - it won't be flagged (and isn't) b/c it only has "eligible" and not "eligible populations in"

tkjones117:
This is very cool. Instead of the hover text reading `This site may be restricted` should we say `Click More Information`? The hover info doesn't provide any more info than the `May be restricted` text and I could imagine people wondering, "What should I do with this info"?

tkjones117:
Oh I see, you need to click on it to get the popup... I totally missed that at first, I thought the hover functionality was the only thing here.

src/CovidAppointmentTable.js
livgust: Yes, sorry. I forgot I snuck it in. I put it at the end of the PR description (first comment). LMK if you'd prefer I remove it and add it separately

src/CovidAppointmentTable.js
livgust: Yeah that's why I did "eligible populations in" - that way it is connoting where you live as opposed to the fact that you're eligible due to phases

tkjones117:
> Oh I see, you need to click on it to get the popup... I totally missed that at first, I thought the hover functionality was the only thing here.

So I wonder if there's a way to make this more intuitive, that the user should click for more info (or remove the popup altogether and just say `Click "More Information" below in the hover text`).

src/CovidAppointmentTable.js
tkjones117: all good! no need to create a separate PR

tkjones117 (approved):
LGTM!

PR #38, opened 2021-02-17
Issue 34/seo

harcod:
Initial attempt at META tags for SEO.  Open Graph (OG) meta for Facebook & Twitter.

image


public/index.html
```diff
@@ -9,10 +9,30 @@
     />
     <meta name="viewport" content="width=device-width, initial-scale=1" />
     <meta name="theme-color" content="#000000" />
-    <meta
-      name="description"
-      content="Make an appointment to get your COVID vaccine"
-    />
+    <meta name="keywords" content="massachusetts, mass, ma, covid, covid-19, vaccine, vaccines, vax, search, sign up, zipcode">
```
j256, line 12: All of these need a self close, right?   <meta ... />

livgust (changes requested):
Hey @harcod, could you add the close `/` for all the meta tags? thanks!

public/index.html
orzechowskid: according to [MDN](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/meta), `<meta>` is a "void element" which means no children and no end tag are allowed.  so `<meta>` with no closing `/>` is correct.

harcod: I added closing tags for all meta

PR #54, opened 2021-02-18
Fixed merge error

j256:
Not 100% sure if we want the "above" or "following".


harcod:
I think "above" was already correct.  
We might want to add the word "**also**" to read...

> We have flagged this site as restricted based on the above information (**also** located under "MORE INFORMATION").

It looks like this on the page.

image

livgust (approved):
thanks!

Issue #22, opened 2021-02-16
Add a link to the GitHub repo under the "About" tab.


JasonD94:
Most sites that have their source on GitHub usually call that out somewhere. Maybe under the "About" page, add something like:

_Source available on [GitHub](https://github.com/livgust/macovidvaccines.com)._

Could also add it to the footer, depends how much you want to call out the source being available.

bernfarr:
If no-one has tackled this yet, I'll work on it.

JasonD94:
That's a good idea @livgust . I didn't realize there is a Slack, so definitely worth pointing out.

bernfarr:
I'm proposing the following changes, I'll create a pull request shortly
From this:
![AboutBefore](https://user-images.githubusercontent.com/2963762/108508474-df74ec80-7289-11eb-95bf-57ecd90f0215.jpg)

To this:
![AboutAfter](https://user-images.githubusercontent.com/2963762/108508483-e26fdd00-7289-11eb-800e-7a5cfb468302.jpg)

There's other work that should be done here. If we're using Material UI, then we should replace `<p>` and `<a>` tags with links. Also probably improve styling use overall. But that's less important that getting these changes in right now.

bernfarr:
Here's what I'm committing shortly. I've removed the direct link to Slack and replaced it with an email link that adds a subject line of **I want to get involved** and a body of **Here's how I can help...**. I also modified the Give Feedback email link to have a subject line of **Site feedback**.

I know there's a leading underscore in the **email the team** link, can't seem to get rid of that.

<img src="https://user-images.githubusercontent.com/2963762/108601075-b24b3b80-7368-11eb-884d-979d584e66f4.png " width="400">

PR #56, opened 2021-02-19
Add github and Slack links to About dialog

bernfarr:
Add links to github repositories and Slack.
#22

Note I also alphabetized the imports, this will be important as code evolves.
Should be have copyright notices in _all_ of the source code?


harcod:
I think we should separate the "About" portion and the "Get Involved" portion of this message.

The "Get Involved" portion is way too nerdy for the general public.  Perhaps this information in a separate dialog could have a link in the footer near  the copyright notice to show that this is an open source project

bernfarr:
I thought about this but don't want to make it overly complicated in a different sense. Most people won't read down this far IMO. If you're browsing on a desktop, an additional dialog wouldn't be challenging. But from a phone?

As it is, the dialog is missing a close X icon in the upper right. I'll have a look around for an about with get involved kind of structure on other sites and see if there's a design we can borrow as an improvement. But in the meantime I'd suggest accepting this.

johnhawkinson:
Screenshots facilitate review.
Before:
Screen Shot 2021-02-19 at 10 06 09


After:
Screen Shot 2021-02-19 at 10 04 47


I would agree we should be a little more clear about whether we're seeking people with software development experience or just general volunteers.

bernfarr:
@johnhawkinson that's the reason for the first sentence. I debated if it was superfluous.

I had the before and after in the issue, happy to also put those in future changes.

Let me know if there's anything else, I'm going to refresh by Reach knowledge wrt Material UI in the meantime.

src/components/Menu.js
```diff
@@ -154,6 +158,36 @@ function AboutDialog(props) {
...
+                    <p>
+                        You can connect with the team on our{" "}
+                        <a
+                            href={"https://macovidvaccines.slack.com/"}
+                            target="_blank"
+                            rel="noreferrer"
+                        >
+                            Slack workspace
+                        </a>.
+                    </p>
```
livgust, line 172: Maybe some disagree, but I'm nervous about making it easy for people to join the Slack workspace. Not that I want it private, but I want to keep it to people who plan on contributing. I'd prefer people request to join via email. Mind updating the copy to reflect that (or making the case for keeping it this way)?

livgust (changes requested):
Copy change re: Slack workspace suggested, otherwise looks great! Thank you!

src/components/Menu.js
harcod: I agree with @livgust .  Otherwise, we will be having random users asking "Where can I get a vaccine?" in Slack

bernfarr:
Committed the following changes and I believe this is ready to merged into master.

I've removed the direct link to Slack and replaced it with an email link that adds a subject line of **I want to get involved** and a body of **Here's how I can help...**. 

I also modified the Give Feedback email link to have a subject line of **Site feedback** and I dropped the capitalization on **F**eedback, it's not a proper noun.

I know there's a leading underscore in the **email the team** link, can't seem to get rid of that.

livgust (approved):

PR #59, opened 2021-02-19
Stale data indicator

livgust:
image
approved UX by Daniel

when more than 8 minutes old, shows timestamp (X:XX AM/PM)
when yesterday, shows yesterday
when older than yesterday, shows date (we should arguably filter these out though, I'll make a new issue)


j256 (approved):

johnhawkinson:
I would push the timestamp to the right margin so it's easy to scan for them.

PR #65, opened 2021-02-20
Updates for cvs

### LINKED ISSUE #62: Updates to match CVS data
1) don't break if no address is sent
2) don't show More Information if there is no More Information (currently we always have at least the address)
3) show totalAvailability if we don't have availability (date-specific) set

livgust:
closes #62 

shows totalAvailability if availability not present/filled out
fix to More Information if address not available


livgust:
tests passing locally for me

PR #68, opened 2021-02-20
Filters out old data

### LINKED ISSUE #60: Filter out entries that are very stale
If we're showing data more than, say, 1 hour old, it's probably not useful and we shouldn't display it at all.

I propose 1 hour but I'm not super tied to that number.

harcod:
Filters out any data that is older than "tooStaleMinutes" which is currently set at "60" (1 hour)
Fixes #60

### DIFF
--- a/src/CovidAppointmentTable.js
+++ b/src/CovidAppointmentTable.js
@@ -16,6 +16,9 @@ import StaleDataIndicator from "./components/StaleDataIndicator";
 import Switch from "@material-ui/core/Switch";
 import Typography from "@material-ui/core/Typography";
 
+// any location with data older than this will not be displayed at all
+export const tooStaleMinutes = 60;  // unit in minutes
+
 export function transformData(data) {
     return data.map((entry, index) => {
         return {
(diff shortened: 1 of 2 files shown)


src/CovidAppointmentTable.js
```diff
@@ -39,9 +39,13 @@ export function sortAndFilterData(
     { sortKey, sortAsc },
     onlyShowAvailable
 ) {
+    const tooStaleMinutes = 60;
+    const oldestGoodTimestamp = new Date() - (tooStaleMinutes * 60 * 1000);
+
     const filteredData = onlyShowAvailable
```
livgust, line 45: I think it would be more readable if you chained `filter`s, so maybe like
```
let filteredData = data.filter((entry) => (entry.timestamp >= oldestGoodTimestamp))
if (onlyShowAvailable) {
    filteredData = filteredData.filter((entry) => entry.hasAppointments);
}
```

src/CovidAppointmentTable.js
```diff
@@ -39,9 +39,13 @@ export function sortAndFilterData(
     { sortKey, sortAsc },
     onlyShowAvailable
 ) {
+    const tooStaleMinutes = 60;
```
livgust, line 42: could you bring this to the top of the file with a comment? for visibility and clarity

livgust (changes requested):
2 nitpicky comments!

src/CovidAppointmentTable.js
harcod: Agreed. Also found logic error for entries with ZERO timestamp.

src/CovidAppointmentTable.js
harcod: Added comment at top of the file. Also commented the code in StaleDataIndictator.js

src/CovidAppointmentTable.js
```diff
@@ -39,10 +42,20 @@ export function sortAndFilterData(
...
     onlyShowAvailable
 ) {
-    const filteredData = onlyShowAvailable
-        ? data.filter((entry) => entry.hasAppointments)
-        : data;
-    const newData = filteredData.sort((a, b) => {
+    // Filter the locations that have "non-stale" data
+    const oldestGoodTimestamp = new Date() - (tooStaleMinutes * 60 * 1000);
+    let filteredData = data.filter(
+        (entry) => (new Date(entry.timestamp).getTime() === 0
```
livgust, line 48: I think it would be clearer to check for null/falsy values here, instead of relying on new Date(null) to be time=0. So
```
({timestamp}) => !timestamp || timestamp >= oldestGoodTimestamp
```

livgust: (switching out `(entry)` for `({timestamp})` because all we care about is `entry.timestamp` but if you prefer to be a little more verbose then I'm cool with keeping `(entry)`)

livgust (changes requested):
Thanks, your changes look good! One last suggestion for clarity & robustness.

harcod:
I had to merge with the changes from PR #69 . 
Please check my work on this.

livgust (approved):
looks good! Verified that Family Practice Group (updated like 2 hrs ago) and another that was updated yesterday were filtered out. Compared production to local server.

PR #69, opened 2021-02-20
Filter changed to only show those locations that have a sign up link.

### LINKED ISSUE #67: Only show appointments should not display sites with no sign up link
Some sites have listed appointment availability, but provide **no sign up link**.  These sites should not be displayed when we are showing "Only show locations with available appointments" as there is no way for the user to obtain an appointment as far as we can tell.

We currently display this advice when prompted about the lack of a sign up link.  We should keep this message as the location will be displayed with the "Only show available appointments" is _unchecked_.

> We were unable to retrieve a sign-up link for this location. It's likely that this location is pulling from a wait list and is not scheduling any new appointments.

EXAMPLE at MAImmunizations:
image

harcod:
Fixes #67

Only shows locations that has a sign up button.

This filters out locations that have a positive number of available slots, but doesn't have any sign up link for the user.

### DIFF
--- a/src/CovidAppointmentTable.js
+++ b/src/CovidAppointmentTable.js
@@ -40,7 +40,7 @@ export function sortAndFilterData(
     onlyShowAvailable
 ) {
     const filteredData = onlyShowAvailable
-        ? data.filter((entry) => hasSignUpLink({ entry }))
+        ? data.filter((entry) => hasSignUpLink(entry))
         : data;
     const newData = filteredData.sort((a, b) => {
         const first = sortAsc ? a[sortKey] : b[sortKey];
--- a/src/CovidAppointmentTable.js
+++ b/src/CovidAppointmentTable.js
@@ -11,7 +11,7 @@ import HelpDialog from "./components/HelpDialog";
 import Loader from "react-loader";
 import MoreInformation from "./components/MoreInformation";
 import React, { useState, useEffect } from "react";
-import SignUpLink, { HasSignUpLink } from "./components/SignUpLink";
+import SignUpLink, { hasSignUpLink } from "./components/SignUpLink";
 import StaleDataIndicator from "./components/StaleDataIndicator";
 import Switch from "@material-ui/core/Switch";
 import Typography from "@material-ui/core/Typography";
@@ -40,7 +40,7 @@ export function sortAndFilterData(
     onlyShowAvailable
 ) {
     const filteredData = onlyShowAvailable
-        ? data.filter((entry) => HasSignUpLink({ entry }))
+        ? data.filter((entry) => hasSignUpLink({ entry }))
         : data;
     const newData = filteredData.sort((a, b) => {
         const first = sortAsc ? a[sortKey] : b[sortKey];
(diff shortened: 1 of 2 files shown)


harcod:
The following is a sample of the type of location that will be filtered out of the default list.

image

j256 (approved):

j256:
Thanks for this.

src/components/SignUpLink.js
```diff
@@ -11,6 +11,26 @@ const useStyles = makeStyles((theme) => ({
     },
 }));
 
+export function HasSignUpLink({ entry }) {
```
livgust, line 14: We don't need to put this in brackets, it can just be hasSignUpLink(entry)

src/CovidAppointmentTable.js
```diff
@@ -11,7 +11,7 @@ import HelpDialog from "./components/HelpDialog";
 import Loader from "react-loader";
 import MoreInformation from "./components/MoreInformation";
 import React, { useState, useEffect } from "react";
-import SignUpLink from "./components/SignUpLink";
+import SignUpLink, { HasSignUpLink } from "./components/SignUpLink";
```
livgust, line 14: do you mind changing it to hasSignUpLink instead of HasSignUpLink? Generally I only do ThisKindOfCamelCase for components, and thisKind for functions and variables. Happy to fight it out though ;)

src/components/SignUpLink.js
```diff
@@ -11,6 +11,26 @@ const useStyles = makeStyles((theme) => ({
...
+        if (entry.signUpLink) {
+            // one sign-up link for all availabilities
+            return true;
+        } else {
+            for (const date in entry.appointmentData) {
+                if (
+                    entry.appointmentData[date].hasAvailability &&
+                    entry.appointmentData[date].signUpLink
+                ) {
+                    // at least one sign-up link found, we're done.
```
livgust, line 25: A+ for cutting out early :)

livgust (changes requested):
A couple quick comments if you don't mind! Tested locally and works as expected 👍

harcod:
This is going to require a careful merge step with PR#68 or vice-versa depending on which one gets accepted first.

src/CovidAppointmentTable.js
```diff
@@ -40,7 +40,7 @@ export function sortAndFilterData(
     onlyShowAvailable
 ) {
     const filteredData = onlyShowAvailable
-        ? data.filter((entry) => entry.hasAppointments)
+        ? data.filter((entry) => hasSignUpLink({ entry }))
```
livgust, line 43: I think this is a bug - we need to remove the {} because then we're sending in `{entry: entry}` as the arg, while the function expects just plain old `entry` now

livgust (changes requested):
One small thing - just need to take out some curly braces!

src/CovidAppointmentTable.js
harcod: Ok.  I guess I need to read up on curly brackets in arguments.

livgust: Curly brackets for args are a short-hand way of saying this:
```function(entry) { //entry has properties "hasAppointments" and "timestamp" that we care about, for example
    const hasAppointments = entry.hasAppointments;
    const timestamp = entry.timestamp;
```
So the above is equivalent to
```
function({hasAppointments, timestamp}) { //now hasAppointments = entry.hasAppointments and timestamp =  entry.timestamp
```

livgust (approved):
Looks good! Tested locally and compared to prod - locations with no sign-up links were filtered out.

PR #70, opened 2021-02-21
Clean up validateDOMNesting console warnings

aceyoung9:
Console shows validateDOMNesting warnings when you open the About/Resources menu options, or when you open the "This site may be restricted" HelpDialogs in the appointment table. This PR addresses that.

Thanks!

### DIFF
--- a/src/components/Menu.js
+++ b/src/components/Menu.js
@@ -119,7 +119,10 @@ function AboutDialog(props) {
                 {"About this Website"}
             </DialogTitle>
             <DialogContent>
-                <DialogContentText id="about-dialog-description">
+                <DialogContentText
+                    id="about-dialog-description"
+                    component="div"
+                >
                     <p>
                         This website was created by{" "}
                         <a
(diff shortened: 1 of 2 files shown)


src/components/Menu.js
```diff
@@ -165,34 +168,38 @@ function ResourcesDialog(props) {
...
-                                    rel="noreferrer"
-                                >
-                                    Vax Finder
-                                </a>{" "}
-                                (state-run)
-                            </li>
-                        </ul>
                     </p>
+
+                    <ul>
```
j256, line 181: Why are these showing up as different?  Are you using different whitespace?  Is there a way to minimize the changes?  Thanks!

aceyoung9: One of the warnings was `Warning: validateDOMNesting(...): <ul> cannot appear as a descendant of <p>.`, so I moved the unordered list out of the paragraph. There are resulting whitespace changes because I am running Prettier on Save.

If you check "Hide whitespace changes" in the settings menu, it will make the change a little easier to read.

image

livgust (approved):

Issue #76, opened 2021-02-21
Remove duplicate "Additional Information:" from some sites


tom-ricci:
Not sure if this was caused by my earlier PR or something else, but on a few locations "Additional Information:" is being written twice

PR #77, opened 2021-02-21
Fixed duplicate Additional Information

tom-ricci:
Issue [#76](https://github.com/livgust/macovidvaccines.com/issues/76)

This PR removes the first occurrence of "Additional Information:" if its at the beginning of the string.

Before:
image
After:
image

Also this PR is for the last 20 commits of my fork instead of just the ones since the last PR, so if there's something I need to do so it only shows the new commits let me know.


livgust (approved):
Confirmed bug is fixed, saw in Reggie Lewis prod vs testing locally. All tests passing.

PR #78, opened 2021-02-21
Future-proof tests from being dependent on the system date and time

MattRC7:
Due to the way our date logic works, when we would run a test around 12pm one of the tests was failing because it wouldn't be the right amount of time apart to display the expected text. Update the tests to minimize the chance of this happening in the future.

Also use some variables so we aren't instantiating date objects a bunch of times in the component.


src/components/StaleDataIndicator.test.js
```diff
@@ -17,17 +17,26 @@ it("shows no indicator if the data is less than 10 minutes old", async () => {
...
             );
         });
 
-        expect(screen.getByText("Last updated 12:00 PM")).toBeTruthy();
+        const timestampDate = new Date(minutesStaleTimestamp);
+        expect(
+            screen.getByText(
+                `Last updated ${
+                    timestampDate.getHours() % 12
+                }:${timestampDate.getMinutes()} PM`
```
livgust, line 37: won't this fail in the morning now? Since it says PM?

src/components/StaleDataIndicator.test.js
```diff
@@ -46,7 +55,7 @@ describe("staleness messaging", () => {
     });
 
     it("shows date if older than yesterday", async () => {
-        const minutesStaleTimestamp = new Date("2021-01-21T00:00");
+        const minutesStaleTimestamp = new Date("2021-01-10T00:00");
```
livgust, line 58: Why? lol

livgust (changes requested):
1 possible bug, 1 nitpick :)

src/components/StaleDataIndicator.test.js
MattRC7: Yep! I'll have to add a ternary here based on whether hours > 12.

src/components/StaleDataIndicator.test.js
MattRC7: Oh this was an artifact from me testing out a mocked Date via `jest.spyOn`. I can revert it if you want

livgust (changes requested):

livgust (changes requested):
see comment re: not padding minutes

src/components/StaleDataIndicator.test.js
```diff
@@ -2,32 +2,43 @@ import { act, render, screen } from "@testing-library/react";
...
                 <StaleDataIndicator
                     timestamp={minutesStaleTimestamp}
-                    staleMinutesOverride={10}
+                    staleMinutesOverride={0}
                 />
             );
         });
 
-        expect(screen.getByText("Last updated 12:00 PM")).toBeTruthy();
+        const timestampDate = new Date(minutesStaleTimestamp);
```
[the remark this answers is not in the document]
MattRC7: Fixed this!

src/components/Availability.js
```diff
@@ -9,22 +9,25 @@ export default function Availability({entry, onlyShowAvailable}) {
...
             // (3) This date has it's own signup link
-            if (entry.appointmentData[date].hasAvailability &&
-                (!onlyShowAvailable || singleSignupLink || entry.appointmentData[date].signUpLink)) {
-
+            if (
+                entry.appointmentData[date].hasAvailability &&
+                (!onlyShowAvailable ||
+                    singleSignupLink ||
+                    entry.appointmentData[date].signUpLink)
+            ) {
```
MattRC7, line 30: This appears to be prettier finding some styling issues that slipped through someone else's commit? I should have an up to date config file

src/components/Availability.js
```diff
@@ -9,22 +9,25 @@ export default function Availability({entry, onlyShowAvailable}) {
     ) {
         return (
             <div>
-                {entry.totalAvailability}
-                {entry.totalAvailability === 1 ? "slot" : "slots"}
+                {`${entry.totalAvailability} ${
+                    entry.totalAvailability === 1 ? "slot" : "slots"
+                }`}
```
MattRC7, line 14: This was breaking a test because it was inserting two different text nodes. I am guessing these were supposed to be one line with a space between them, eg. "45 slots"? If not lemme know and I'll revert

MattRC7:
As for why I didn't install `momentjs`, check the [project status](https://momentjs.com/docs/#/-project-status/). They have some alternatives recommended. I can create an issue in the Project if you'd like, to convert the app to use one of those recommendations instead of the rather flaky `Date` API that the browser/node provides.

src/components/StaleDataIndicator.test.js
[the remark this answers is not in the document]
livgust: verified this works on the hour too (0 minutes)

livgust (approved):

PR #82, opened 2021-02-23
Refactored NoAppointmentsAlert

### LINKED ISSUE #79: Unit Test Fails on "No Appointments Found"
With the introduction new code in PR #75, there is a failed test in CovidAppointmentTable.test.js at line 76 (test named "when the api endpoint can not be reached")

There in this case there are 2 "role=alert" divs in the file.  Is this causing the failure?

harcod:
Fixed #79 

Refactored NoAppointmentsAlert
Added real UI for the ErrorMessageAlert
replaced previous check for an 'alert' role in the error message

### DIFF
--- a/src/CovidAppointmentTable.js
+++ b/src/CovidAppointmentTable.js
@@ -2,7 +2,7 @@
 
 import { makeStyles } from "@material-ui/core";
 import Alert from "@material-ui/lab/Alert";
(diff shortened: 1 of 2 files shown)


harcod:
**Not a real problem!** This is just how it looks if there ever happens to be an error.

image

livgust (approved):

PR #84, opened 2021-02-24
Test front-end with cached data

### LINKED ISSUE #83: Showing Availability when there is no Signup Link
We are showing availabiity on dates that don't have a sign up link as shown in this screen shot:

image

The _Sign Up_ button in this case goes with the 1 slot on 03/04.  Note that there is no date selector in front of the sign up link because there is only a single day with a sign up link.

We should not display the availability of 8 slots on 02/23 as there is no way for the user to signup for this slot.

harcod:
Fixes #83 

Example: 
The Sign Up button in this case goes with the 1 slot on 03/04. Note that there is no date selector in front of the sign up link (in **BEFORE**) because there is only a single day with a sign up link.

In **AFTER**, we no longer display the availability of 8 slots on 02/23 as there is no way for the user to signup for this slot.

**BEFORE**:

**AFTER**:

image


harcod:
I have a "test/devtest.json" file for this case, but I don't want to add it to the repo.  I can email the file to you if you want to test against it.

harcod:
When I don't have a "test/devtest.json" file, then I get the following message:

```
Compiled with warnings.

./src/test/devtest.json
Module build failed: Error: ENOENT: no such file or directory, open '/home/.../macovidvaccines.com/src/test/devtest.json'

```

It works in development, but is unsettling.  Hopefully this doesn't break production.

harcod:
I separated this PR from PR#85 so the UI can be fixed without installing this new testing harness.

livgust:
@harcod is this still a draft or is ready to be evaluated for a merge?

harcod:
@liv It still has the "compiled with warnings" problem.  I'll take another look to see if I can remove that problem.

harcod:
Closing this PR.  There is a newer PR #145 that implements this feature.

PR #85, opened 2021-02-24
Hide availability when no sign up link

### LINKED ISSUE #83: Showing Availability when there is no Signup Link

harcod:
Fixes #83 

Example: 
The Sign Up button in this case goes with the 1 slot on 03/04. Note that there is no date selector in front of the sign up link (in **BEFORE**) because there is only a single day with a sign up link.

In **AFTER**, we no longer display the availability of 8 slots on 02/23 as there is no way for the user to signup for this slot.

**BEFORE**:

**AFTER**:

image


harcod:
I separated this PR from PR #84 
This means that there is no good way for you to see the test case.
The test file that I used is available at `s3://ma-covid-vaccine/data-2021-02-23T2253Z.json`

livgust (changes requested):
One nitpicky comment; I'll test this weekend!

livgust (approved):

Issue #36, opened 2021-02-17
Show total available appointments across all sites


JasonD94:
Could be interesting & useful to display a sum of all the available appointments across all sites at the bottom of the website.

Something like:

All Sites:
2/18/2021: #### appointments
2/19/2021: #### appointments
         Total: #### appointments 

This could go on a separate page too, something like "additional info".

PR #91, opened 2021-02-27
Display total available slots

ramon-h:
Displaying total available appointments for a location as part of https://github.com/livgust/macovidvaccines.com/issues/36.

I had to pull in a https://github.com/livgust/macovidvaccines.com/commit/e777424d262689f77ae4d35720d606d43995b271 from `master` with some changes that weren't in `develop` yet.


livgust (changes requested):

src/components/Availability.js
```diff
@@ -37,6 +51,11 @@ export default function Availability({ entry }) {
                             }`}
                         </div>
                     ))}
+                    {totalAvailableSlots > 1 && availableSlots.length > 1 && (
```
ramon-h, line 54: I updated this to `availableSlots.length > 1` because I noticed that if there's only 1 slot available, then the total is duplicative.
In my last commit, I had `> 0` so we get this:
Screen Shot 2021-02-28 at 7 12 23 PM

And now with `> 1` we only see this case:
Screen Shot 2021-02-28 at 7 12 15 PM

src/components/Availability.js
```diff
@@ -3,6 +3,16 @@ import HelpDialog from "./HelpDialog";
 export default function Availability({ entry }) {
     if (!entry.hasAppointments) {
         return <div>No availability.</div>;
+    } else if (
+        entry.totalAvailability &&
```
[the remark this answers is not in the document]
ramon-h: That's a good a good question. I don't see that data, but that change came from here https://github.com/livgust/macovidvaccines.com/commit/e777424d262689f77ae4d35720d606d43995b271
I pulled it in to avoid a merge conflicts.

src/components/Availability.js
```diff
@@ -26,6 +36,10 @@ export default function Availability({ entry }) {
                 </div>
             );
         } else {
+            const totalAavailableSlots = availableSlots.reduce(
```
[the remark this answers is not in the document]
ramon-h: Thanks. This should be fixed now.

src/components/Availability.js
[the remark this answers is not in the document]
livgust: LOL of course it's something I wrote 🤦 thanks!

livgust (commented):
Hey @ramon-h , thanks for doing this. The functionality is good. I want to run it by UX and get their feedback real quick on the design; I'll get back to you soon.

livgust:
Hey @ramon-h, could we move Total available to the top of the list (before date-specific information) and bold the entire line?

ramon-h:
Here's what it looks like now. I used an h6 for the total. Also, just to confirm, the total does not show when there is only 1 "row" of slots since the total would match the other row.

Screen Shot 2021-03-03 at 8 59 14 PM

src/components/Availability.js
```diff
@@ -36,8 +37,17 @@ export default function Availability({ entry }) {
...
             );
         } else {
+            const totalAvailableSlots = availableSlots.reduce(
+                (total, slot) => total + slot.numberAvailableAppointments,
+                0
+            );
             return (
                 <div>
+                    {totalAvailableSlots > 1 && availableSlots.length > 1 && (
+                        <Typography variant="h6">
```
MattRC7, line 47: For accessibility reasons, we want to avoid skipping header levels. Unless there is an `h5` somewhere higher up the page, we should change this. If you need the `h6` default styles, create a class that imitates this.

src/components/Availability.test.js
```diff
@@ -49,3 +49,40 @@ it("shows total slots if availability has no content", async () => {
...
+                        "10/12/2021": {
+                            hasAvailability: true,
+                            numberAvailableAppointments: 35,
+                        },
+                    },
+                }}
+            />
+        );
+    });
+    expect(screen.queryByText("Total available: 70 slots")).toBeTruthy();
```
MattRC7, line 73: Prefer `expect(getByText()).toBeVisible()` over this pattern, as it will return more descriptive errors on failure.

src/components/Availability.test.js
```diff
@@ -49,3 +49,40 @@ it("shows total slots if availability has no content", async () => {
...
+        render(
+            <Availability
+                entry={{
+                    hasAppointments: false,
+                    totalAvailability: 0,
+                }}
+            />
+        );
+    });
+    expect(screen.queryByText("Total available")).toBeFalsy();
```
MattRC7, line 87: Same as above.

MattRC7 (commented):
Left a couple small suggestions!

src/components/Availability.js
livgust: Hey @ramon-h , actually UX wants it just bold so you can remove the heading :)

ramon-h: Just pushed the change. I had noticed the usage of Material but not `makeStyles`.

src/components/Availability.test.js
ramon-h: Thanks for the suggestion. Just pushed the change.

src/components/Availability.test.js
ramon-h: I tried making the same change here, but it didn't work out quite as easily. I was searching for the absence of text and `getByText` throws an error when the text doesn't exist so I followed the `toThrow` example here: https://jestjs.io/docs/en/expect#tothrowerror

ramon-h:
This should be all set.

livgust:
tests pass. verified that total available appears at the top if more than one date has availability.

PR #93, opened 2021-03-01
Merge master into develop

harcod:
Second attempt to merge master into develop.

This has the "drawer-based" UI for filters.  There are only 2 filters.

1. Show Available Appointments
2. Don't show stale data [ no UI for this filter ]

The VaxTypeFilter is commented out in the file (and it wouldn't work correctly even if uncommented)

This is intended to be a stepping off point for adding additional filters.


harcod:
@liv OK. I think this is going to be correct this time.  Let me know if you have any questions or concerns at all.

.circleci/config.yml
```diff
@@ -1,24 +1,24 @@
 # config file for circleci
```
livgust, line 1: Could you update `prettier` to do 2 spaces for yaml/yml files? If it's not already in `master` for this repo, it's in `master` for the other one so you can copy-paste :)

src/components/FilterPanel.js
```diff
@@ -1,15 +1,21 @@
+/*eslint no-unused-vars: ["error", { "varsIgnorePattern": "[iI]gnored" }]*/
+
 import React, { useEffect, useState } from "react";
 import { makeStyles } from "@material-ui/core";
 import { useTheme } from "@material-ui/core/styles";
 import useMediaQuery from "@material-ui/core/useMediaQuery";
 import FormControl from "@material-ui/core/FormControl";
-import FormLabel from "@material-ui/core/FormLabel";
+//import FormLabel from "@material-ui/core/FormLabel";
```
livgust, line 8: do you know why this is commented out by chance?

livgust (changes requested):
2 questions for you! Then I'll approve

.circleci/config.yml
harcod: Can I just delete the "**.prettier.json**" file and add a "**.prettierrc**" file from the `covid-vaccine-scrapers`?

src/components/FilterPanel.js
harcod: It's not being used right now.  I'll can just delete it in the next update.

src/components/FilterPanel.js
```diff
@@ -147,21 +148,26 @@ export default function FilterPanel(props) {
             include: Array.apply(null, Array(vaxTypes.length)).map((d) => true),
         });
     }, [data]);
-
+*/
     useEffect(() => {
         onChange({
+            showStaleData: (d) => {
```
livgust, line 154: if we're never actually going to give the user the chance to filter stale or non-stale data, we should remove this. I think it's slowing the logic down because every time you mess with a filter you're re-evaluating everything to also check if it's stale again.

livgust (changes requested):
I was only able to look a little bit but I noticed 2 things:
1) clicking the "Has Available Appointments" checkbox is REALLY slow. I think part of it is because (left a comment above) whenever it's clicked we're also re-filtering out stale entries. That's added computation we don't need.
2) If you deselect "Has Available Appointments" and then resize the window such that the display changes (if the side bar was docked before, now it's not, or vice versa) when you open the filters again the checkbox is checked even though you deselected it. The state of the checkbox should refer to the same React Hook ("useState" variable) regardless of visuals.

I'm not sure how much you're planning on addressing within this PR so let me know if you want me to ignore this and have it be dealt with later (but obviously before pushing to `master`)

src/components/FilterPanel.js
harcod: ok. I'll try to move this to a higher level of filtering.

PR #95, opened 2021-03-02
Filters out "too stale" data immediately

harcod:
Filters out all of the stale data immediately at the transformData call.

FYI. This reduced the data set from 132 entries to 95.


harcod:
Made the same update as PR #96 for the develop branch.

livgust (approved):

PR #104, opened 2021-03-03
chore: run prettier

tkjones117:
Make it pretty so hooks pass.


johnhawkinson (approved):
Yup, this is fine.
As a style note, I would put a lot more clarity in the commit message (e.g. filename), even you only keep it to one line.
Probably I'm the only one, tho!

MattRC7 (approved):
Seconding this approval! @livgust sorry to bug, but this is causing commit hooks to fail, could you take a look tonight?

livgust (approved):

PR #107, opened 2021-03-03
add alert for March 4

tkjones117:
In preparation for tomorrow's rush, let's get an alert ready to push.


tkjones117:
Screen Shot 2021-03-03 at 11 40 59 AM

This is the same alert we used last week, except I removed the timestamp.

livgust (approved):
I added a timer so it doesn't appear until 6 AM tomorrow morning (eastern)

PR #109, opened 2021-03-03
Synced checkboxes in both FilterPanels.

harcod:
_@livgust wrote in a previous review:_
> I was only able to look a little bit but I noticed 2 things:
> 
> 1. clicking the "Has Available Appointments" checkbox is REALLY slow. I think part of it is because (left a comment above) whenever it's clicked we're also re-filtering out stale entries. That's added computation we don't need.
> 2. If you deselect "Has Available Appointments" and then resize the window such that the display changes (if the side bar was docked before, now it's not, or vice versa) when you open the filters again the checkbox is checked even though you deselected it. The state of the checkbox should refer to the same React Hook ("useState" variable) regardless of visuals.
> 
> I'm not sure how much you're planning on addressing within this PR so let me know if you want me to ignore this and have it be dealt with later (but obviously before pushing to `master`)

(1) was fixed in another PR and has already been merged.
This fixes part (2) above.  The two checkboxes are synced with a common "useState" variable.

I'm a little disappointed with the structure of this fix because it creates a state linkage between the App and FilterPanel because there are TWO FilterPanels (one permanent, one temporary for mobile).  As we add more filters to the FilterPanel, we are going to have to add more state to the App.  _This is just not very modular._ There is probably a better way to accomplish this, but I'm currently at a loss on how to do that.

It still appears to be calling the filtering code twice when a checkbox is toggled.  I have no idea why this is the case. Not the end of the world, but it is "added computation we don't need".

### DIFF
--- a/src/App.js
+++ b/src/App.js
@@ -1,3 +1,5 @@
+/*eslint no-unused-vars: ["error", { "varsIgnorePattern": "[iI]gnored" }]*/
+
 import Alert from "@material-ui/lab/Alert";
 import AlertTitle from "@material-ui/lab/AlertTitle";
 import ArrowBack from "@material-ui/icons/ArrowBack";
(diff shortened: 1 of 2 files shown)


livgust:
I think there shouldn't be 2 FilterPanels - that whole section (the two <Hidden>s) should be their own component that manages state between the mobile and desktop views. Merging into `develop` anyway :)

PR #110, opened 2021-03-03
Lighthouse

ramon-h:
This PR sets up lighthouse-ci so you can generate reports locally and configures a GitHub Action to run them with commits. I disabled a few of the assertions so that it wouldn't generate an errors as part of this. This also fixes a few of the minor issues that came up in the reports.


livgust:
Hi @ramon-h ! Can you tell me more about what lighthouse-ci does here? I'm not familiar :)

ramon-h:
> Hi @ramon-h ! Can you tell me more about what lighthouse-ci does here? I'm not familiar :)

Sure. So [Lighthouse](https://developers.google.com/web/tools/lighthouse) is an opensource tool from Google to analyze a website and it checks for a variety of best practices around performance, accessibility, SEO, etc. The tool is built into the Chrome browser and you can run it manually here:
Screen Shot 2021-03-04 at 9 03 25 AM

lighthouse-ci takes that and lets you run it from the command line, which means you can integrate it into a build process. It's possible to have it trigger build failures, as a way to enforce standards. So I know that axe is running, but that does require someone to remember to look at the console, which doesn't always happen. This would ensure that every push to GitHub would trigger a report and the results are saved as an artifact in GitHub.

Screen Shot 2021-03-04 at 10 34 40 AM

ramon-h:
I thought I'd point out that there's a potential performance improvement by adding compression to the endpoint you're using to provide the data.

Screen Shot 2021-03-04 at 6 58 08 PM

livgust (approved):

PR #111, opened 2021-03-04
Configure react-i18n for future translations

MattRC7:
## Summary
This PR installs and configures `i18next` and `react-i18next`, and adds a single English translation for the Website main content header, as a preemptive step to translating the text of the website into other languages.

## Caveats / Discussion Topics
I've implemented the translations file structure as a set of JSON files within `/src`. This is acceptable for a small set of translations. If those files grow in size, we may want to move all files besides a fallback file into `/public` and install the [http backend plugin](https://github.com/i18next/i18next-http-backend) to fetch them asynchronously, or decide to pay for the Locize service backend.

I've also opted to use the `useTranslation()` hook. This is one of several options, and I'm interested in hearing what other's think.


MattRC7:
Confirmed that both the page title and the "# slot(s)" text work for English:

image

livgust (approved):

PR #119, opened 2021-03-06
StateEligiblity.js: Eligibility string enhancements

johnhawkinson:
3 related but independent enhancements for StateEligibility.js:
* Refactor to editable strings/array/object
* Use PeopleIcon instead of CheckIcon
* Update teachers' URL

### DIFF
--- a/src/components/StateEligibility.js
+++ b/src/components/StateEligibility.js
@@ -55,7 +55,7 @@ const criteriaGroups = [
         title: "Eligible to sign up starting March 11: ",
         list: [
             [
-                "https://www.mass.gov/info-details/massachusetts-covid-19-vaccination-phases",
+                "https://www.mass.gov/news/baker-polito-administration-announces-k-12-educators-child-care-workers-and-k-12-school-staff",
                 "Pre-K to 12 teachers and staff, early educators, and child care workers",
             ],
         ],
--- a/src/components/StateEligibility.js
+++ b/src/components/StateEligibility.js
@@ -56,7 +56,7 @@ const criteriaGroups = [
         list: [
             [
                 "https://www.mass.gov/news/baker-polito-administration-announces-k-12-educators-child-care-workers-and-k-12-school-staff",
-                "Pre-K to 12 teachers and staff, early educators, and child care workers",
+                "K-12 educators, child care workers and K-12 school staff",
             ],
         ],
     },
--- a/src/components/StateEligibility.js
+++ b/src/components/StateEligibility.js
@@ -24,16 +24,43 @@ const useStyles = makeStyles((theme) => ({
     },
 }));
 
-function EligibilityGroupItem(props) {
-    return (
-        <ListItem>
-            <ListItemIcon>
-                <CheckIcon />
-            </ListItemIcon>
-            {props.children}
-        </ListItem>
-    );
-}
+const criteriaGroups = [
+    {
+        title: null,
+        list: [
+            [
+                "https://www.mass.gov/info-details/covid-19-vaccinations-for-people-ages-65-and-older",
+                "Individuals age 65 and older",
+            ],
+            [
+                "https://www.mass.gov/info-details/covid-19-vaccinations-for-individuals-with-certain-medical-conditions",
+                "Individuals with two or more of certain medical conditions",
+            ],
+            [
+                "https://www.mass.gov/info-details/covid-19-vaccinations-for-senior-housing-settings",
+                "Residents and staff of low-income and affordable senior housing",
+            ],
+            [
+                "https://www.mass.gov/info-details/covid-19-vaccinations-for-people-ages-75-and-older",
+                "Individuals age 75 and older",
+            ],
+            [
+                "https://www.mass.gov/info-details/massachusetts-covid-19-vaccination-phases#phase-1-",
+                "People in Phase 1 (healthcare, nursing homes, etc.)",
+            ],
+        ],
+    },
+    /* TODO - remove the following div after March 11, and update link to be appropriate link from https://www.mass.gov/covid-19-vaccine */
+    {
+        title: "Eligible to sign up starting March 11: ",
+        list: [
+            [
+                "https://www.mass.gov/info-details/massachusetts-covid-19-vaccination-phases",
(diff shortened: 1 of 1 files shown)


johnhawkinson:
After:
Screen Shot 2021-03-05 at 23 14 10


Before:
Screen Shot 2021-03-05 at 23 14 15

(Please ignore the grey difference, that was a rendering artifact as I was taking the screenshot)

johnhawkinson:
(Sorry about those force-pushes, I need to read the test logs more closely.)
Also, it turns out there's a `next week` icon (not clear if people would understand it, but starting tomorrow it'll be "this week" so wrong) as well as some medical/hospital -type icons that could also be used. Food for minor thoughts:

Screen Shot 2021-03-06 at 00 12 18
Screen Shot 2021-03-06 at 00 12 28
Screen Shot 2021-03-06 at 00 12 45

harcod:
> Pre-K to 12 teachers and staff, early educators, and child care workers

The press release that you link to defines this group as:
**K-12 educators, child care workers and K-12 school staff**

I didn't see "Pre-K" anywhere in that press release.

johnhawkinson:
> I didn't see "Pre-K" anywhere in that press release.

Good eye, Rob!

@tkjones117, this was your language from cf528d1c0d95db7d5b54e9bb3b224b09bc026f07, where did it come from?

As a practical matter, it's hard for me to imagine a "pre-K educator" who does not qualify as a "child care worker," but government has been known to leave my imagination in the dust, and where we can use the official language we should do so when there is no reason to depart from it.

This shouldn't block this PR, but I'm happy to add such a wording fix if we have concensus, and @harcod's formatting/casing looks right to me:
> **K-12 educators, child care workers and K-12 school staff**

tkjones117:
Got the wording from @danieljae ’s PR feedback

harcod (approved):

livgust:
@tkjones117 @johnhawkinson has UX approved these changes ?

johnhawkinson:
> @tkjones117 @johnhawkinson has UX approved these changes ?

Not exactly, but not un-exactly. **With regard to icons,** as explained in 581077ebc64de0fe21958f9644e39eae81f05c58, 

> As noted by me and confirmed by Hal, the CheckIcon had an affordance
for completion, the implication that you're supposed to do all the
checkable items. Although it's not seriously contemplated that anyone
would actually *think* that, it's the wrong application for a checkbox
icon.

> Although we originally discussed using a bullet icon, material UI
doesn't seem to have a good choice -- there is
RadioButtonUncheckedIcon but it's a fairly large circle.

> Instead, let's use icons that actually represent groups of people,
which is what we are listing anyhow. That means...PeopleIcon!

So, I wasn't aware we had a formal "UX approval" process but in conversation, Hal agreed we should move away from the checkmarks. Given the realities of the other material-ui choices, PeopleIcon seems proper, but that's why Daniel is requested to review and Hal is tagged (because Hal isn't a Contributor so can't be requested to review).

**With regard to the rest**, I don't think UX needs to weigh in on the maintainability of strings, or even necessarily on the rewording of language where it's clear we have inadvertently diverged from the State's messaging.

livgust:
Thanks for the update, I agree. I'll also put this in the #ux channel for visibility.

src/components/StateEligibility.js
```diff
@@ -48,71 +75,44 @@ export default function StateEligibility() {
...
-                            <a
-                                href="https://www.mass.gov/info-details/massachusetts-covid-19-vaccination-phases"
-                                rel="noreferrer"
-                                target="_blank"
-                            >
-                                Pre-K to 12 teachers and staff, early educators,
-                                and child care workers
-                            </a>
-                        </EligibilityGroupItem>
+                        {criteriaGroups.map((g, idx) => {
```
livgust, line 78: nothing to change right now but I prefer using slightly more descriptive variable names, unless it's like a two-liner or something.

src/components/StateEligibility.js
```diff
@@ -48,71 +75,44 @@ export default function StateEligibility() {
...
+                                            <>
+                                                <br />
+                                                <br />
+                                                <div>{g.title}</div>
+                                                <br />
+                                            </>
+                                        ) : null
+                                    }
+                                    {
+                                        //
```
livgust, line 97: remove please!

src/components/StateEligibility.js
```diff
@@ -48,71 +75,44 @@ export default function StateEligibility() {
...
+                                // However the interaction with fragments is unclear, so ensure
+                                // title fragments and listitems are unique, as they are siblings.
+                                <React.Fragment key={"title" + idx}>
+                                    {
+                                        // We don't display a title for the first group, for some reason...
+                                        g.title ? (
+                                            <>
+                                                <br />
+                                                <br />
+                                                <div>{g.title}</div>
```
livgust, line 91: nitpick I'm not requiring you fix: instead of relying on the correct number of breaks it'd be cleaner to style the div with padding.

livgust (changes requested):
quick remove the empty comment row and you're good :)

src/components/StateEligibility.js
johnhawkinson: This PR attempted to make no visible changes in the final DOM other. Those `<br/>`s were there before and I agree they're probably wrong but someone wiser than I approved them in the past.

src/components/StateEligibility.js
johnhawkinson: You're quite right it started life as a 1-liner but didn't  get revised.
Fixed in https://github.com/johnhawkinson/macovidvaccines.com/commit/ebd313eec28d349812283ea1645028941263b1ad
which will be overwritten in the PR by the force-push for clean history for a merge commit.

src/components/StateEligibility.js
johnhawkinson: Gone. (If anyone is wondering, it was there to make things prettier).
Fixed in https://github.com/johnhawkinson/macovidvaccines.com/commit/aeb550866e3be40612ccaaaa86f5b35e3d8a1cb2
which will be overwritten in the PR by the force-push for clean history for a merge commit.

livgust (approved):

Issue #118, opened 2021-03-06
Out of state zip has unintended behavior


livgust:
Replication steps:
- enter zip in Massachusetts
- notice mile distance appears
- enter zip out of state (like 12345, which is in Schenectady)
- notice same milage appears as with MA zip

harcod:
**Reason**: `us-Zips` is not returning coordinates for 12345.

harcod:
Local test case: **02634** (Centerville, MA) is a ZIP Code for PO Boxes and thus not in the ZCTAs from the Census.  Here is a [list of ZIP Codes for Centerville](https://www.zip-codes.com/city/ma-centerville.asp). Note that 02634 is a zero population area of PO Boxes.

Looks like we might need another way to get this information.

PR #120, opened 2021-03-06
Invalid (or Unknown) ZIP Codes sort by location.

harcod:
The  `us-Zips` library is based on ZCTAs (ZIP Code Tabulation Areas) and does not contain every ZIP Code.  Basically, just ZIP Codes with populations attached.  This commonly excludes PO Box zip codes and "unique" ZIP Codes like 12345.  I think that USPS probably just mapped 12345 as unused because of the common pattern and just assigned to Schenectady, NY because they other 123xx codes.

Code was updated to use the default "location sort order" if the ZIP code is invalid or doesn't have coordinates.  It was previously defaulting to the last computation at reported in Issue #118

### DIFF
--- a/src/components/FilterPanel.js
+++ b/src/components/FilterPanel.js
@@ -13,7 +13,7 @@ import Grid from "@material-ui/core/Grid";
 import Typography from "@material-ui/core/Typography";
 import Container from "@material-ui/core/Container";
 import { hasSignUpLink } from "./SignUpLink";
-import { getDistance } from "geolib";
+import { convertDistance, getDistance } from "geolib";
 import { setSortBy } from "../services/appointmentData.service";
 import Cookies from "universal-cookie";
 
@@ -224,14 +224,19 @@ export default function FilterPanel(props) {
                         const myCoordinates = usZips[zipCodeFilter.zipCode];
                         if (myCoordinates) {
                             setSortBy("miles");
-                            const metersPerMile = 1609.34;
                             d.miles = Math.round(
-                                getDistance(myCoordinates, d.coordinates, 1) /
-                                    metersPerMile
+                                convertDistance(
+                                    getDistance(
+                                        myCoordinates,
+                                        d.coordinates,
+                                        1
+                                    ),
+                                    "mi"
+                                )
                             );
 
                             // Is the location within the range specified?
-                            return d.miles < zipCodeFilter.miles;
+                            return d.miles <= zipCodeFilter.miles;
                         }
                     }
                 }
--- a/src/components/FilterPanel.js
+++ b/src/components/FilterPanel.js
@@ -221,9 +221,9 @@ export default function FilterPanel(props) {
                 if (d) {
                     const zipValid = zipCodeFilter.zipCode.match(/\d{5}/);
                     if (zipValid) {
-                        setSortBy("miles");
                         const myCoordinates = usZips[zipCodeFilter.zipCode];
                         if (myCoordinates) {
+                            setSortBy("miles");
                             const metersPerMile = 1609.34;
                             d.miles = Math.round(
                                 getDistance(myCoordinates, d.coordinates, 1) /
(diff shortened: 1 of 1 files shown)


livgust:
FYI 12345 is actually Schenectady! It's for the GE campus there. My husband went to college nearby and it's local trivia :)

src/components/FilterPanel.js
```diff
@@ -221,9 +221,9 @@ export default function FilterPanel(props) {
                 if (d) {
                     const zipValid = zipCodeFilter.zipCode.match(/\d{5}/);
                     if (zipValid) {
-                        setSortBy("miles");
                         const myCoordinates = usZips[zipCodeFilter.zipCode];
                         if (myCoordinates) {
+                            setSortBy("miles");
```
johnhawkinson, line 226: I know this is out of scope of this change, but:

 `geolib`'s `getDistance()` function says it's 3rd arg is "accuracy (in meters). By default the accuracy is 1 meter. … You can set it higher to have the result rounded to the next value that is divisible by your chosen accuracy."

~Why are we rounding ourselves instead of using the built-in rounding function?~

OK, we're not rounding we're _converting_.

Why are we converting ourselves with a hardcoded physical constant (what if the earth accelerates and we redefine the relationship between meters and feet because the US is left behind the metric system?) instead of using `geolib.convertDistance(value, 'mi')`?

src/components/FilterPanel.js
```diff
@@ -233,11 +233,12 @@ export default function FilterPanel(props) {
                             // Is the location within the range specified?
                             return d.miles < zipCodeFilter.miles;
```
johnhawkinson, line 234: Also out of scope:

If I ask for locations within 10 miles, I think I want `loc <= 10 miles` not `loc < 10 miles`?

src/components/FilterPanel.js
```diff
@@ -233,11 +233,12 @@ export default function FilterPanel(props) {
...
                         }
-                    } else {
-                        setSortBy("location");
                     }
                 }
-                return true; // include all for now.
+                // No zipcode was provided or
+                // was unable to find coordinates for the specified zipcode
+                setSortBy("location");
```
johnhawkinson, line 240: So, if I type in "treehouse," the filter doesn't show me all the vax sites near my treehouse.
This is unreasonable, but what can you do?

Nevertheless, there's no indication or feedback that it even _tried_. I'm not sure what to do about that, but it seems like there should be … something.

harcod:
If we want to support 12345, then we are going to have to find another library or data source for this information.  This PR only ignores the ZIP Codes that don't have coordinates available via us-Zips.

If you type in 12345 now, you will just get a list sorted by location name as if you didn't type in a ZIP Code at all.  The previous code was switching to "sort by miles" from the _last valid ZIP Code_ that was computed.

johnhawkinson (approved):
This reads and tests fine.
Approved.

Looking at this code for real for the first time, I have some comments that are out-of-scope to this PR. I thought about leaving them as github comments on the prior commits to this file, e.g. d118a5be3cc8e5efb8a9530fb569e36c2a662cb8, or opening a new Issue, and it just seemed easier to toss them in here. They are not blockers and hardly even relevant and I'll accept the criticism that they don't belong here at all (but then…_where_?)

src/components/FilterPanel.js
harcod: I didn't see "convertDistance"... good idea.  I still have to round the miles though because the accuracy is in meters.

src/components/FilterPanel.js
harcod: I agree, but was unsure _where_ to display such an error message.  This is also a problem in the interim case of while you are entering a ZIP Code and stop at "020"

johnhawkinson (approved):
Looks and tests fine.
(well, prettier came in with a galactic miter saw and cut all your lines into little strips 1 attoparsec wide, but what can you do?)

src/components/FilterPanel.js
harcod: @liv Maybe you can incorporate this idea while you are working on the FilterPanel.   Since the panel is non-modal, we need someway to signify that the ZIP Code is **not valid**.  This might be because we don't have a **long/lat** for it, or maybe its not well-formed **/\d{5}/**

livgust (approved):

PR #123, opened 2021-03-06
UX: LocationCard visual changes

harcod:
There have been a number of items suggested in the #ux channel.  This PR address a few of them.

1. Keep the city near the location
2. There was no space separating MA and ZIP Code in MoreInformation.... was ",MA01760".  Now ",MA 01760"
3. Changed the size of the "stale icon" to match the "restriction icon"
4. Changed "**Important Eligibility Notice**" to "**Information about eligibility**" as recommended in the [#ux channel by Daniel](https://macovidvaccines.slack.com/archives/C01N7C2G2KF/p1614795429083700)

**BEFORE:**
image

**AFTER:** _(the data wasn't stale on the 2nd screen shot! The icon is the same size as the restriction)_
image

### DIFF
--- a/src/CovidAppointmentTable.js
+++ b/src/CovidAppointmentTable.js
@@ -193,7 +193,7 @@ function RestrictionNotifier({ entry }) {
                 }
             >
                 <Typography className={classes.restrictionWarning}>
-                    Important Eligibility Notice
+                    Information about eligibility
                 </Typography>
             </HelpDialog>
         );
@@ -213,13 +213,13 @@ function LocationCard({ entry, className, onlyShowAvailable }) {
                     }
                     subheader={
                         <>
-                            <RestrictionNotifier entry={entry} />
                             <div>
                                 {entry.city}{" "}
                                 {sortedByMiles()
                                     ? `(${entry.miles} miles)`
                                     : ""}
                             </div>
+                            <RestrictionNotifier entry={entry} />
                             <StaleDataIndicator timestamp={entry.timestamp} />
                         </>
                     }
--- a/src/components/MoreInformation.js
+++ b/src/components/MoreInformation.js
@@ -50,8 +50,7 @@ export default function MoreInformation({ entry }) {
                                 rel="noreferrer"
                                 href={googleMapsLink}
                             >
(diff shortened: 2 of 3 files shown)


harcod:
Map link:
**BEFORE**
image

AFTER:
image

src/CovidAppointmentTable.js
```diff
@@ -193,7 +193,7 @@ function RestrictionNotifier({ entry }) {
                 }
             >
                 <Typography className={classes.restrictionWarning}>
-                    Important Eligibility Notice
+                    Information about eligibility
```
johnhawkinson, line 196: Was this change from "Important Eligibility Notice" to "Information about eligibility" intended?
I didn't see it explained.

johnhawkinson (commented):
With respect to the the first visual mock, and I don't mean that this is important, but I had to put something in the "review" text, so here we are:

There is a subtle visual change going from
  BIG, RED, grey, darkgrey
to
  BIG, grey, RED, (darkgrey?)
that perhaps suggests that if we wanted to retain the original "look" (err…"type color," by which I mean density and not hue).
It's sort of like we from iambic to trochaic [-/ to /-]? (well, maybe more spondaic to trochaic? [-- to -/])

Perhaps an adjustment in the leading (err, line height/spacing), might be called for.
probs not worth it.

src/CovidAppointmentTable.js
harcod: It was recommend in the [#ux channel by Daniel](https://macovidvaccines.slack.com/archives/C01N7C2G2KF/p1614795429083700)

johnhawkinson: 'k. It should appear in the commit message when this is merged to `master`, whether by way of being in the summary of this PR or by being in an additional individual commit you further push to the PR now.

johnhawkinson (approved):

src/CovidAppointmentTable.js
harcod: I edited the PR comment at the top.

livgust (approved):

PR #125, opened 2021-03-06
Only use Massachusetts zipcodes for better performance

johnhawkinson:
We save 374K, or 60% of the gzipped project size, by only including Massachusetts zipcodes and not the entire country.
That matters for mobile.

It brings us down to, as reported by `npm build`:

```
File sizes after gzip:

  230.26 KB (-382.98 KB)  build/static/js/2.abb59050.chunk.js
  16.04 KB (+8.39 KB)     build/static/js/main.95eb2b80.chunk.js
  1.41 KB (+1 B)          build/static/js/3.52f81af8.chunk.js
  1.18 KB                 build/static/js/runtime-main.a4b75426.js
```

This is my first foray into npm scripts called from `package.json`, so be noisy if you dislike the choices I've made or the filenames or the locations. I would not have placed generated files under `src/` but `require()` and `import` seem to need that.

Apparenlly `devDependencies` don't make it into `package-lock.json` ... err, except they did after I forced an uninstall of `us-zips`, I...don't get that.

To get there, we

* FilterPanel.js: Prefer generic name `zipcodeData``

(unnecessary commit hairsplitting for the sake of future merger that will never happen)

* prettier `.prettierrc.json`
* `.prettierignore`: sort
* prettier: Ignored generated JSON

`JSON.stringify(Object)` is not pretty, and satisfying Prettier requires more than just `JSON.stringify(Object,null,1)`, so give up and ignore the generated JSON. While we're here, fix prettier config issues.

* Script to generate `ma-zips.json`
* package.json: Generate ma-zips
* FilterPanel.js: Switch to MA zipcodes only.


package.json
```diff
@@ -52,6 +53,7 @@
     },
     "devDependencies": {
         "husky": "^5.0.9",
-        "prettier": "^2.2.1"
+        "prettier": "^2.2.1",
+        "zipcodes-perogi": "^1.4.201908"
```
livgust, line 57: this shouldn't be in both regular dependencies AND devdependencies

scripts/generate-ma-zips.js
```diff
@@ -0,0 +1,39 @@
+// Generate a pared down list of Massachusetts zipcodes with their
+// longitude and latitude.
+
+const fs = require("fs");
+const zipcodes = require("zipcodes-perogi");
+
+const MAZipcodes = zipcodes.lookupByState("MA");
+
+// 'zipcodes-perogi' gives us this form, which is more than we need:
```
livgust, line 9: suuuuper nit but could you use /* instead of the individual line comments?

package.json
livgust: After actually looking at your code I think we can just leave it in devDeps, right?

livgust (changes requested):
stupid nitpicks!

package.json
johnhawkinson: It should not have been in `dependencies`! I don't know how that happened!
Done in 008cbd79f636bdac5f61cce0283251b5c12e7930

scripts/generate-ma-zips.js
johnhawkinson: Sure, but…why? What's the rule? If that's a project style, shouldn't prettier enforce it?

I haven't heard anyone argue for mulitline /* comments in ages, but I've definitely heard good arguments for // (namely, ease of using simple editor tools to insert and remove them, and not having to deal with the computational complexity of nesting if you block-comment-out a region or worse, reverse it.)

livgust: Hm, I don't think prettier has a setting for it. But I don't care enough so if you're hesitant to change it we can leave it. I like it for readability personally - lots of `//` lines feels cluttered to me 🤷

livgust (approved):

scripts/generate-ma-zips.js
johnhawkinson: > if you're hesitant to change it we can leave it.

Nah, it's changed: 6f3f931c6bac439f4e5d19a9dfab0e8e1c959002

I just like to know why: I always assume I was one of those kids who went around asking "Why is the sky blue?" long past "Rayleigh scattering" and so I do want to know what the thinking is. (Sounds like the answer is "mental model of neatness that is formed by habit but not grounded in clearly-articulated principles," which is fine.)

Of course, as a philosophical matter, style opinions that are not written down and not enforced by the code formatter are maybe not the best. And where they seem to differ with others considered defaults (but not preferences), that's extra friction.

While I'm here, I force-pushed over your merge commit (2a00ba5d637bd268d3247e8fe8070e9054385133), I don't think it's necessary to have in here and think it will either lead to a less clean merge later or I'll want to rebase the branch before merging. (But this PR is definitely a candidate for a squash merge, if not for the prettier stuff, so cleanliness of branch history isn't a big deal.) I can bring it back if you want tho.

harcod (approved):
Looks good!

PR #129, opened 2021-03-07
Refactor AlertBanner out of App.js

harcod:
Created a separate component for the site banner.  This alert appears at the top of the site (usually on Thursday mornings).

This should eliminate having to deal with merge conflicts on App.js when we are just changing the banner text.

Fixes a spacing issue with css that was previously accomplished with a "<br />" after the alert

Added new functionality:
**startDate** and **endDate** so we can have the alert appear and vanish at predetermined intervals.  There is a sample alert from last week that is commented out in the file.

### DIFF
--- a/src/components/AlertBanner.js
+++ b/src/components/AlertBanner.js
@@ -9,30 +9,52 @@ const useStyles = makeStyles((theme) => ({
     },
 }));
 
+/* There are four severity levels each has its own icon built-in
+ *  error   - red
+ *  warning - orange (** this is what you should probably use)
+ *  info    - blue
+ *  success - green
+ */
+/* The "-05:00" at the end of the dates adjust from EST to UTC
+ * NOTE: DST begins on 3/14/2021 when the offset becomes "-04:00"
+ * NOTE: EST begins on 11/7/2021 when the offset becomes "-05:00"
+ */
 const alerts = [
-    /*    {
-        startDate: "2021-03-04T06:00:00-05:00",
-        endDate: "2021-03-04T09:00:00-05:00",
-        severity: "warning",
+    /*
+    {
+        startDate: "2021-03-04T06:00:00-05:00", // current timezone offset is at the end
+        endDate: "2021-03-24T09:00:00-05:00",
+        severity: "warning", // one of "error", "info", "success", "warning"
         title: "Thursday, March 4",
         contents:
             "Due to high demand, the MA vaccination websites are experiencing technical difficulties. Once the issues are resolved, their locations will appear on this website.",
-    },*/
+    },
+*/
 ];
 
 export default function AlertBanner() {
     const classes = useStyles();
 
+    // No alerts? Then we're done!
+    if (!alerts.length) {
+        return false;
+    }
+
+    // Filter all of the timely alerts
     const now = new Date();
-    return alerts.map((alert) => {
-        return new Date(alert.startDate) <= now &&
-            now < new Date(alert.endDate) ? (
+    const timelyAlerts = alerts.filter((alert) => {
+        return (
+            new Date(alert.startDate) <= now && now < new Date(alert.endDate)
+        );
+    });
+
+    // Output an Alert for each of the remaining timely alerts
+    return timelyAlerts.map((alert) => {
+        return (
             <Alert severity={alert.severity} className={classes.alertBanner}>
                 <AlertTitle>{alert.title}</AlertTitle>
                 {alert.contents}
             </Alert>
-        ) : (
-            <></>
         );
     });
 }
--- a/src/components/AlertBanner.js
+++ b/src/components/AlertBanner.js
@@ -12,7 +12,7 @@ const useStyles = makeStyles((theme) => ({
 const alerts = [
     /*    {
         startDate: "2021-03-04T06:00:00-05:00",
-        endDate: "2021-03-24T09:00:00-05:00",
+        endDate: "2021-03-04T09:00:00-05:00",
         severity: "warning",
         title: "Thursday, March 4",
         contents:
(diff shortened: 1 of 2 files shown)


src/components/AlertBanner.js
```diff
@@ -0,0 +1,38 @@
...
+        contents:
+            "Due to high demand, the MA vaccination websites are experiencing technical difficulties. Once the issues are resolved, their locations will appear on this website.",
+    },*/
+];
+
+export default function AlertBanner() {
+    const classes = useStyles();
+
+    const now = new Date();
+    return alerts.map((alert) => {
```
livgust, line 27: combining a map with a filter would be clearer IMO and would remove the empty <></>s that we're conditionally returning.

livgust (changes requested):
Thanks for making this refactor! One suggestion, LMK if you think it would be smart to do.

src/components/AlertBanner.js
```diff
@@ -0,0 +1,38 @@
...
+        return new Date(alert.startDate) <= now &&
+            now < new Date(alert.endDate) ? (
```
johnhawkinson, line 29: This would be more readable as

```
        const valid =
            new Date(alert.startDate) <= now && now < new Date(alert.endDate);
        return valid ? (
```

src/components/AlertBanner.js
```diff
@@ -0,0 +1,38 @@
...
+            <Alert severity={alert.severity} className={classes.alertBanner}>
+                <AlertTitle>{alert.title}</AlertTitle>
+                {alert.contents}
+            </Alert>
+        ) : (
+            <></>
+        );
```
johnhawkinson, line 36: `( <></> )` can be replaced with `false`.

src/components/AlertBanner.js
```diff
@@ -0,0 +1,38 @@
...
+        marginBottom: theme.spacing(2),
+        display: "flex",
+    },
+}));
+
+const alerts = [
+    /*    {
+        startDate: "2021-03-04T06:00:00-05:00",
+        endDate: "2021-03-04T09:00:00-05:00",
+        severity: "warning",
```
johnhawkinson, line 16: If we anticipate this being editing in a rush by inexperienced React developers, perhaps a coment summarizing the choices for severity, which seem to be: success, info, warning, error.

johnhawkinson (commented):
This seems fine, I have only small nits.

Part of me wants to offer a long discursion about how merge conflicts shouldn't be happening and how that is reflective of a different problem, but the rest of me note that is emperically incorrect and this definitely going to help. ("It's all well and good that they shouldn't be happening, but they are.")

src/components/AlertBanner.js
harcod: adding @livgust  filter suggestion will take care of this

src/components/AlertBanner.js
harcod: adding @livgust  filter suggestion will take care of this

harcod:
I think I resolved all of the comments from your review.  Thanks for the suggestions.

src/components/AlertBanner.js
```diff
@@ -9,30 +9,52 @@ const useStyles = makeStyles((theme) => ({
...
+/* There are four severity levels each has its own icon built-in
+ *  error   - red
+ *  warning - orange (** this is what you should probably use)
+ *  info    - blue
+ *  success - green
+ */
+/* The "-05:00" at the end of the dates adjust from EST to UTC
+ * NOTE: DST begins on 3/14/2021 when the offset becomes "-04:00"
+ * NOTE: EST begins on 11/7/2021 when the offset becomes "-05:00"
+ */
```
johnhawkinson, line 21: The JS `Date()` API is so dumb. It is tempting to suggest the user should not be specifying the timezone and should simply specify a date in Eastern time, but I agree with your implicit conclusion that that is just way too much effort for this. It just makes me upset.

src/components/AlertBanner.js
```diff
@@ -9,30 +9,52 @@ const useStyles = makeStyles((theme) => ({
...
+*/
 ];
 
 export default function AlertBanner() {
     const classes = useStyles();
 
+    // No alerts? Then we're done!
+    if (!alerts.length) {
+        return false;
+    }
```
johnhawkinson, line 41: This is unnecessary?

johnhawkinson (approved):
Seems fine. If you wanna go crazy you can look at what I wrote but it's OK as-is!

src/components/AlertBanner.js
harcod: We want our times to be Eastern Time, but then we have to deal with what timezone the server is in.

I thought about trying to solve the timezone problem and I **would** if we were crashing a rover on Mars.  (I'll bet that Mars doesn't even have timezones _yet_).

src/components/AlertBanner.js
harcod: I know, but it made it perfectly clear that nothing was going to be emitted in the _zero_ case.

src/components/AlertBanner.js
johnhawkinson: It does not have zones _per se_ insofar as there are not geographical bounds within which rovers (or people) crossing them change time standards in an agreed-upon fashion, but there are many different time standards that are used on Mars.

See https://en.wikipedia.org/wiki/Timekeeping_on_Mars#Summary for a summary of the time standards used by the 7 US landers, which are variants of Local Mean Solar Time and Local True Solar Time. And then you can read the entire article for more time zones used with-respect-to (but not actually on) Mars. It seems somewhat US centric, I don't know what _Hope_ and _Tianwen-1_ use?

tkjones117 (approved):
Nice!

PR #133, opened 2021-03-08
UX: Updates Eligibility, MoreInformation, Cards

harcod:
Here is the big [Slack discussion](https://macovidvaccines.slack.com/archives/C01N7C2G2KF/p1615137139112600)

**New Look:**

**CLOSED Accordions:**
image

**EXPANDED Accordions**
image

### DIFF
--- a/src/components/StateEligibility.js
+++ b/src/components/StateEligibility.js
@@ -3,6 +3,7 @@ import Accordion from "@material-ui/core/Accordion";
 import AccordionSummary from "@material-ui/core/AccordionSummary";
 import AccordionDetails from "@material-ui/core/AccordionDetails";
 import ExpandMoreIcon from "@material-ui/icons/ExpandMore";
+import Link from "@material-ui/core/Link";
 import List from "@material-ui/core/List";
 import ListItem from "@material-ui/core/ListItem";
 import ListItemIcon from "@material-ui/core/ListItemIcon";
@@ -10,10 +11,11 @@ import ListItemText from "@material-ui/core/ListItemText";
 import PeopleIcon from "@material-ui/icons/People";
 import { makeStyles } from "@material-ui/core/styles";
 import { Typography } from "@material-ui/core";
+import grey from "@material-ui/core/colors/grey";
 
 const useStyles = makeStyles((theme) => ({
     container: {
-        marginBottom: theme.spacing(2),
+        marginBottom: theme.spacing(1),
         display: "flex",
         justifyContent: "center",
     },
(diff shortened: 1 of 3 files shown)


danieljae (approved):

harcod:
Hal commented on Slack...
Re: links. I don’t have a rule or any user data to say why I don’t like the full lines being linked text. It just doesn’t feel right. The “learn more” link seemed like a good compromise, but I think @jhawk has a good point about linking it this way “Individuals age 65 and older”.

In retrospect, linking “learn more” is like linking “click here”, which one should never do. So I was wrong. Let’s use “Individuals age 65 and older” instead. (The website uses underlines, which Slack doesn’t, btw. I like underlined links.)

harcod:
Updated linked text segments. Underlining links everywhere!

image

harcod:
Tightened up the icon spacing.  Added timestamps to the data to automatically handle changes to the phases (ala K-12 on March 11th)

image

On March 11th, this will automatically be displayed as:

image

harcod:
Added the newest link to Mass.gov for K-12.
Refactored the markup code to provide a link in the title line.

src/components/StateEligibility.js
```diff
@@ -22,61 +24,124 @@ const useStyles = makeStyles((theme) => ({
...
+    },
+    accordionExpanded: {
+        backgroundColor: grey[300],
+    },
+    listGroup: {
+        paddingTop: theme.spacing(2),
+        paddingBottom: theme.spacing(1),
+    },
+    listItem: {
+        padding: 0, // theme.spacing(0),
```
tkjones117, line 44: Did you mean to leave this comment?

src/components/StateEligibility.js
```diff
@@ -118,3 +172,67 @@ export default function StateEligibility() {
...
+        </ListItem>
+    );
+}
+
+/*
+ * @param {string} text
+ * @param {string} link
+ * @param {string} variant
+ */
+function MarkupLink({ text, link, variant }) {
```
tkjones117, line 213: This is a clever way to simplify the info we need to provide when creating the links!

tkjones117: I didn't verify the regex is 100% correct but I trust it from the output visual result

src/components/StateEligibility.js
```diff
@@ -22,61 +24,124 @@ const useStyles = makeStyles((theme) => ({
...
+
+            {
+                link:
+                    "https://www.mass.gov/info-details/massachusetts-covid-19-vaccination-phases#phase-1-",
+                text: "People in [Phase 1] (healthcare, nursing homes, etc.)",
+                color: "primary",
+            },
         ],
     },
     /* TODO - remove the following div after March 11, and update link to be appropriate link from https://www.mass.gov/covid-19-vaccine */
```
tkjones117, line 103: I love that it will automatically switch over at midnight. And then we can still remove the below code tmrw.

tkjones117 (approved):

livgust (commented):
This was merged before I finished my review. No biggie, I just had some small comments. Leaving them here for posterity.

PR #134, opened 2021-03-09
fixes build by using require instead of import as recommended per the…

iann:
fixes build by using require instead of import as recommended per the @axe-core/react docs


livgust (approved):
ran
```
npx cross-env CI=true npm run build
```
with no errors, built successfully.

PR #137, opened 2021-03-09
Actually update filter design

livgust:


livgust:
image
received UX approval already

harcod (approved):
This is looking great!
I was expecting to be able to press "Enter" on my keyboard to press the "Update List" button.  I approved the PR anyway since this could be debated in #usability

hshubin:
I agree that pressing Return or Enter should be like clicking the submit button. Fix it now if it's easy, or in an update.

harcod:
Bug Report.  LocationCard needs to have the "onlyShowAvailable" boolean passed to it.  This appears to have been lost during the big "develop" to "master" merge.

See CovidAppointmentTable.js, Line 125

                        <LocationCard
                            entry={entry}
                            className={classes.cardBox}
                            onlyShowAvailable={onlyShowAvailable}
                            key={getSiteId(entry)}
                            showMiles={sortBy === "miles"}
                        />

iann:
I think it works well. 

One thing I didn't realize immediately was that you needed to click update again when you select another radio button distance for the list to update. I thought there was mention of graying out the button when it's set and re-enabling it when they are changes to update? Probably not a big deal.

iann (approved):

PR #141, opened 2021-03-12
Preregistration Banner

harcod:
Site Banner & Menu>Resources

(and for some reason I could commit code because of 'jest' FAILS.
Fixed one. Skipped the other.

### DIFF
--- a/src/components/FilterPanel/index.js
+++ b/src/components/FilterPanel/index.js
@@ -116,10 +116,13 @@ export default function FilterPanelParent({
                                 className={classes.mobileButton}
                                 onClick={() => setFilters(inProgressFilters)}
                                 disabled={
-                                    inProgressFilters.filterByZipCode.zipCode &&
-                                    !isZipValid(
+                                    !!(
                                         inProgressFilters.filterByZipCode
-                                            .zipCode
+                                            .zipCode &&
+                                        !isZipValid(
+                                            inProgressFilters.filterByZipCode
+                                                .zipCode
+                                        )
                                     )
                                 }
                                 type="submit"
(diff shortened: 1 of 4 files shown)


harcod:
image

image

harcod:
Live Preview at https://covid.harcod.com/PR141/

johnhawkinson (approved):
Looks good, ran it, passes tests, we should get that out fast!

src/components/FilterPanel/index.js
```diff
@@ -116,10 +116,13 @@ export default function FilterPanelParent({
                                 className={classes.mobileButton}
                                 onClick={() => setFilters(inProgressFilters)}
                                 disabled={
-                                    inProgressFilters.filterByZipCode.zipCode &&
-                                    !isZipValid(
+                                    !!(
```
tkjones117, line 119: @harcod did you mean to do this? idk how this is related to the change at hand

harcod: (Unrelated but...) Yes.  It was required to pass a jest test.  It wouldn't let me commit!!

harcod: This change forced it to be a 'boolean' that jest was complaining about.

tkjones117 (approved):
Approving b/c this is marked as Urgent and Rob will make sure the following get addressed:
- fix the cause of needing to skip the jest test
- `Warning: Each child in a list should have a unique "key" prop.`

harcod:
Added [issue 142](https://github.com/livgust/macovidvaccines.com/issues/142) to address this

> Approving b/c this is marked as Urgent and Rob will make sure the following get addressed:
> 
> * fix the cause of needing to skip the jest test
> * `Warning: Each child in a list should have a unique "key" prop.`

PR #144, opened 2021-03-13
No longer skips test

### LINKED ISSUE #142: App.test.js fails two tests
● Console

    console.error
      Warning: Each child in a list should have a unique "key" prop. See https://reactjs.org/link/warning-keys for more information.
          at AlertBanner (/home/rob/source/macovid/livgust/macovidvaccines.com/src/components/AlertBanner.js:51:21)
          at div
          at Grid (/home/rob/source/macovid/livgust/macovidvaccines.com/node_modules/@material-ui/core/Grid/Grid.js:241:35)
          at WithStyles(ForwardRef(Grid)) (/home/rob/source/macovid/livgust/macovidvaccines.com/node_modules/@material-ui/styles/withStyles/withStyles.js:67:31)
          at div
          at Grid (/home/rob/source/macovid/livgust/macovidvaccines.com/node_modules/@material-ui/core/Grid/Grid.js:241:35)
          at WithStyles(ForwardRef(Grid)) (/home/rob/source/macovid/livgust/macovidvaccines.com/node_modules/@material-ui/styles/withStyles/withStyles.js:67:31)
          at div
          at Grid (/home/rob/source/macovid/livgust/macovidvaccines.com/node_modules/@material-ui/core/Grid/Grid.js:241:35)
          at WithStyles(ForwardRef(Grid)) (/home/rob/source/macovid/livgust/macovidvaccines.com/node_modules/@material-ui/styles/withStyles/withStyles.js:67:31)
          at main
          at MainComponent (/home/rob/source/macovid/livgust/macovidvaccines.com/src/App.js:75:21)
          at ThemeProvider (/home/rob/source/macovid/livgust/macovidvaccines.com/node_modules/@material-ui/styles/ThemeProvider/ThemeProvider.js:48:24)
          at App

      at printWarning (node_modules/react-dom/cjs/react-dom.development.js:67:30)
      at error (node_modules/react-dom/cjs/react-dom.development.js:43:5)
      at warnForMissingKey (node_modules/react-dom/cjs/react-dom.development.js:13135:5)
      at warnOnInvalidKey (node_modules/react-dom/cjs/react-dom.development.js:13602:11)
      at reconcileChildrenArray (node_modules/react-dom/cjs/react-dom.development.js:13651:21)
      at reconcileChildFibers (node_modules/react-dom/cjs/react-dom.development.js:14125:14)
      at reconcileChildren (node_modules/react-dom/cjs/react-dom.development.js:16990:28)
      at mountIndeterminateComponent (node_modules/react-dom/cjs/react-dom.development.js:17890:5)

    console.error
      Warning: Each child in a list should have a unique "key" prop.
      
      Check the render method of `StateEligibility`. See https://reactjs.org/link/warning-keys for more information.
          at CriterionItem (/home/rob/source/macovid/livgust/macovidvaccines.com/src/components/StateEligibility.js:188:26)
          at StateEligibility (/home/rob/source/macovid/livgust/macovidvaccines.com/src/components/StateEligibility.js:120:21)
          at div
          at Grid (/home/rob/source/macovid/livgust/macovidvaccines.com/node_modules/@material-ui/core/Grid/Grid.js:241:35)
          at WithStyles(ForwardRef(Grid)) (/home/rob/source/macovid/livgust/macovidvaccines.com/node_modules/@material-ui/styles/withStyles/withStyles.js:67:31)
          at div
          at Grid (/home/rob/source/macovid/livgust/macovidvaccines.com/node_modules/@material-ui/core/Grid/Grid.js:241:35)
          at WithStyles(ForwardRef(Grid)) (/home/rob/source/macovid/livgust/macovidvaccines.com/node_modules/@material-ui/styles/withStyles/withStyles.js:67:31)
          at div
          at Grid (/home/rob/source/macovid/livgust/macovidvaccines.com/node_modules/@material-ui/core/Grid/Grid.js:241:35)
          at WithStyles(ForwardRef(Grid)) (/home/rob/source/macovid/livgust/macovidvaccines.com/node_modules/@material-ui/styles/withStyles/withStyles.js:67:31)
          at main
          at MainComponent (/home/rob/source/macovid/livgust/macovidvaccines.com/src/App.js:75:21)
          at ThemeProvider (/home/rob/source/macovid/livgust/macovidvaccines.com/node_modules/@material-ui/styles/ThemeProvider/ThemeProvider.js:48:24)
          at App

      148 |                                 // However the interaction with fragments is unclear, so ensure
      149 |                                 // title fragments and listitems are unique, as they are siblings.
    > 150 |                                 <React.Fragment key={"title" + index}>
          |                                 ^
      151 |                                     {group.title ? (
      152 |                                         // We don't display a title for the first
      153 |                                         // group, for some reason...

      at printWarning (node_modules/react/cjs/react-jsx-dev-runtime.development.js:117:30)
      at error (node_modules/react/cjs/react-jsx-dev-runtime.development.js:93:5)
      at validateExplicitKey (node_modules/react/cjs/react-jsx-dev-runtime.development.js:986:5)
      at validateChildKeys (node_modules/react/cjs/react-jsx-dev-runtime.development.js:1013:11)
      at jsxWithValidation (node_modules/react/cjs/react-jsx-dev-runtime.development.js:1174:15)
      at map (src/components/StateEligibility.js:150:33)
          at Array.map (<anonymous>)
      at StateEligibility (src/components/StateEligibility.js:144:41)

harcod:
Closes #142

**Test**: "when the api endpoint can not be reached"

**Before**: Was failing because there are multiple "role=alert" on the screen when the site's AlertBanner is displayed.

**Now**: This test checks for the existence of "Unexpected Internal Error" which is the title of the ErrorMessageAlert

### DIFF
--- a/src/App.test.js
+++ b/src/App.test.js
@@ -77,12 +77,12 @@ describe("the App component", function () {
             );
         });
 
-        test.skip("it displays an error message", async function () {
+        test("it displays an error message", async function () {
             await act(async function () {
                 render(<App />);
             });
 
-            expect(screen.getByRole("alert")).toBeInTheDocument();
+            expect(screen.getByText("Unexpected Internal Error"));
         });
     });
 });
--- a/src/App.test.js
+++ b/src/App.test.js
@@ -44,8 +44,8 @@ describe("the App component", function () {
             });
 
             // FYI: There two buttons (one for mobile, one for desktop)
-            await screen.getAllByTestId("onlyShowAvailable")[0].click();
-            await screen.getByTestId("UpdateListDesktop").click();
+            await screen.getByTestId("availability-checkbox").click();
+            await screen.getByTestId("update-list-button").click();
 
             expect(await screen.findAllByRole("listitem")).toHaveLength(3);
         });
--- a/src/App.test.js
+++ b/src/App.test.js
@@ -44,12 +44,8 @@ describe("the App component", function () {
             });
 
             // FYI: There two buttons (one for mobile, one for desktop)
-            await screen
-                .getAllByText(
-                    "Only show locations with available appointments"
-                )[0]
-                .click();
-            await screen.getAllByText("Update List")[0].click();
+            await screen.getAllByTestId("onlyShowAvailable")[0].click();
+            await screen.getByTestId("UpdateListDesktop").click();
 
             expect(await screen.findAllByRole("listitem")).toHaveLength(3);
         });
--- a/src/App.test.js
+++ b/src/App.test.js
@@ -38,13 +38,18 @@ describe("the App component", function () {
             expect(await screen.findAllByRole("listitem")).toHaveLength(2);
         });
 
-        //TODO: UPDATE
-        test.skip("disabling the filter shows all appointment cards", async function () {
+        test("disabling the filter shows all appointment cards", async function () {
             await act(async function () {
                 render(<App />);
             });
 
-            (await screen.findByLabelText("switch")).click();
+            // FYI: There two buttons (one for mobile, one for desktop)
+            await screen
+                .getAllByText(
+                    "Only show locations with available appointments"
+                )[0]
+                .click();
+            await screen.getAllByText("Update List")[0].click();
 
             expect(await screen.findAllByRole("listitem")).toHaveLength(3);
         });
--- a/src/components/FilterPanel/RadiusFilter.js
+++ b/src/components/FilterPanel/RadiusFilter.js
@@ -61,7 +61,11 @@ export default function RadiusFilter({ value, onChange }) {
 }
 
 export function isWithinRadius(item, zipCode, miles) {
-    if (isZipValid(zipCode)) {
+    if (
+        isZipValid(zipCode) &&
+        item.coordinates.latitude &&
+        item.coordinates.longitude
+    ) {
         const myCoordinates = zipcodeData[zipCode];
         item.miles = Math.round(
             convertDistance(
--- a/src/services/appointmentData.service.js
+++ b/src/services/appointmentData.service.js
@@ -4,7 +4,7 @@ import { isWithinRadius } from "../components/FilterPanel/RadiusFilter";
 const dayjs = require("dayjs");
 
 // any location with data older than this will not be displayed at all
-export const tooStaleMinutes = 60; // unit in minutes
+export let tooStaleMinutes = 60; // unit in minutes
 
 export function transformData(data) {
     const ourDateFormat = "M/D/YY"; // 3/2
@@ -44,6 +44,7 @@ export function transformData(data) {
         return !d.timestamp || d.timestamp >= oldestGoodTimestamp;
     });
 }
+
 export function sortData(data, sortKey) {
     const newData = data.sort((a, b) => {
         const first = a[sortKey];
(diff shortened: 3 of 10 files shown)


johnhawkinson (approved):
I am probably not competent to review this.

It certainly works and the test passes and it is better than skipping the test.
But it also feels like it's not consistent with the way the test used to run, so one of the two is/was wrong, or at least philosophically different. And I am not sure how it should really go.

These are the alerts we now have:
```
      --------------------------------------------------
      alert:

      Name "":
      <div
        class="MuiPaper-root MuiAlert-root MuiAlert-standardSuccess makeStyles-alertBanner-168 MuiPaper-elevation0"
        role="alert"
      />

      Name "":
      <div
        class="MuiPaper-root MuiAlert-root MuiAlert-standardError MuiPaper-elevation0"
        role="alert"
      />
```

It is tempting to suggest that we should continue to use `getByRole("alert")` and restrict it by class (`MuiAlert-standardError`) to get as close to where we were before, but `getByRole()` doesn't offer us that flexibility. 

The use of `getByText()` to look for the for the boilerplate "Unexpected Internal Error" just feels generic — what other conditions aside from the error we have generated would produce that? Should we be also testing its absence without mocking in the network error? But that's infeasible without accessing the network which is undesirable.

Or if we're using `getByText()` maybe we should be looking for "network error" or maybe we should be changing that string to something a little more specific. But it feels like we're getting out the thing we put in, which is a little circular.

So I dunno. Someone who knows this stuff better should say. But it's definitely better than where we are today.

johnhawkinson:
> The use of `getByText()` to look for the for the boilerplate "Unexpected Internal Error" just feels generic — what other conditions aside from the error we have generated would produce that? Should we be also testing its absence without mocking in the network error? ~But that's infeasible without accessing the network which is undesirable.~ **I guess we can just mock in fake JSON data if we wanted to do that, so it _is_ feasible.**

In particular, should we be doing _this_?: https://github.com/livgust/macovidvaccines.com/commit/ef496d3d37ab3183feb29f9a2ae753512dd225f6

(I genuinely am not sure if this is a pointless thing to do.)

harcod:
There was another skipped test because of recent changes to filtering.
The test was updated to click on the "_Only Show locations with available appointments_" checkbox, and the "_Update List_" button.

src/App.test.js
```diff
@@ -38,13 +38,18 @@ describe("the App component", function () {
...
             });
 
-            (await screen.findByLabelText("switch")).click();
+            // FYI: There two buttons (one for mobile, one for desktop)
+            await screen
+                .getAllByText(
+                    "Only show locations with available appointments"
+                )[0]
+                .click();
+            await screen.getAllByText("Update List")[0].click();
```
johnhawkinson, line 52: Is this going to break with internationalization? Should we not assign an ID and check for the ID instead?

harcod: I changed the test to use `data-testid`.

src/App.test.js
```diff
@@ -44,12 +44,8 @@ describe("the App component", function () {
...
 
             // FYI: There two buttons (one for mobile, one for desktop)
-            await screen
-                .getAllByText(
-                    "Only show locations with available appointments"
-                )[0]
-                .click();
-            await screen.getAllByText("Update List")[0].click();
+            await screen.getAllByTestId("onlyShowAvailable")[0].click();
+            await screen.getByTestId("UpdateListDesktop").click();
```
johnhawkinson, line 48: Why the difference?

livgust (changes requested):

src/components/FilterPanel/RadiusFilter.js
```diff
@@ -61,7 +61,11 @@ export default function RadiusFilter({ value, onChange }) {
 }
 
 export function isWithinRadius(item, zipCode, miles) {
-    if (isZipValid(zipCode)) {
+    if (
+        isZipValid(zipCode) &&
+        item.coordinates.latitude &&
```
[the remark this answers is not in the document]
harcod: I looked into this (and it is `?.` btw).  For reference, this is [optional chaining notation](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Optional_chaining)

Not really needed here because the object is guaranteed to have a `coordinates` key after going through `transformData` in `appointmentData.service.js`.  The problem is the source data doesn't have `entry.latitude` or `entry.longitude` -- thus the added test.

I was seeing the case in archived data set where `item.coordinates == { latitude: undefined, longitude: undefined }`

src/services/appointmentData.service.js
```diff
@@ -73,9 +74,43 @@ export function filterData(data, { filterByAvailable, filterByZipCode }) {
...
+                console.log("Data: View Source");
+                cachedTransform = transformData(testData.body.results);
+            }
+        } catch (err) {
+            // if the file doesn't exist, just get the prod file
+            console.log("using production file: err= " + err);
+        }
+
+        if (cachedTransform) {
+            tooStaleMinutes = 60 * 24 * 365 * 25; // 25 years! otherwise, you might not see anything
```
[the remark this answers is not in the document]
harcod: Great idea! However, I'm going to address that in PR 145.  I didn't actually mean for this file to be changed at all in this PR (oops).  I'm going to roll back to the previous version of `appointmentData.service.js`

src/App.test.js
harcod: I update the two `FilterPanel`s to have unique `data-testid` for the checkboxes and buttons.

livgust (approved):

PR #145, opened 2021-03-13
Testing front-end with local data

harcod:
Note: This only runs in development mode. `if (process.env.NODE_ENV === "development")`

This is a testing branch to get data from a local file instead of the production file.
It will read a file called "**test/devtest.json**" in the src directory.
If the file doesn't exist, then it fetches the production data.

You can obtain a cached file using a cmd line:
`aws s3 cp s3://ma-covid-vaccine/data-2021-02-23T2253Z.json src/test/devtest.json`


harcod:
Now uses the timestamp in the test file (if available) as a time reference instead of `now = new Date()`.

This will make the archived data look like it did on the date it was saved.  Otherwise, data would be marked as **tooStale** and removed, or **a little stale** and labeled in the UI.  

If there is no timestamp available, then we use a _now date_ in the past so that everything is in the _future_ and not stale.  This isn't exactly how it would have looked on the day in question, but the alternative is to display "No Appointments Found" as all of the data would be stale.

src/services/appointmentData.service.js
```diff
@@ -54,28 +66,71 @@ export function sortData(data, sortKey) {
...
-        }
-        return true;
+        );
     });
 }
 
 export function getAppointmentData() {
+    let testDataTransformed = null;
+
+    if (process.env.NODE_ENV === "development") {
```
livgust, line 86: We should change this to != production

In reality it's only ever set when we're in prod, as opposed to always set to dev otherwise

src/services/appointmentData.service.js
```diff
@@ -54,28 +66,71 @@ export function sortData(data, sortKey) {
...
+                testDataTransformed = transformData(testData.results);
+            }
+        } catch (err) {
+            // if the file doesn't exist, just get the prod file
+            testDataTransformed = null;
+        }
+    }
+
+    // If there is not any test d
+    if (!testDataTransformed) {
```
livgust, line 125: Will this use our endpoint if there's no test data? That would be ideal!

(I'd test but I'm trying to do this from my phone while I feed the baby)

src/CovidAppointmentTable.js
```diff
@@ -11,73 +11,11 @@ import HelpDialog from "./components/HelpDialog";
...
-import SignUpLink, { hasSignUpLink } from "./components/SignUpLink";
+import SignUpLink from "./components/SignUpLink";
 import { sortData } from "./services/appointmentData.service";
 import StaleDataIndicator from "./components/StaleDataIndicator";
 import Typography from "@material-ui/core/Typography";
 
-const dayjs = require("dayjs");
-
-// any location with data older than this will not be displayed at all
-export const tooStaleMinutes = 60; // unit in minutes
```
livgust, line 22: Maybe I'm not understanding but why are we removing all this code? We still want it to run in prod at least, right?

livgust (changes requested):
A couple Qs. Trying to review from my phone so I apologize for not quite grasping the whole picture here!

src/services/appointmentData.service.js
harcod: The tests were failing.  Now I think this happens because I changed "now" to be the timestamp on any test data file that might be in the way.

src/services/appointmentData.service.js
harcod: Exactly!  If you want to run the code with _live data_ then you just have to delete the "test/devtest.json" file (or have a malformed one)

src/CovidAppointmentTable.js
harcod: All of this code was refactored in a previous PR to `appointmentData.service.js`.  I think it must have been accidentally reinserted into this file during a merge.

src/services/appointmentData.service.js
harcod: Finally settled on adding an environment variable to `.env.test`

src/services/appointmentData.service.js
```diff
@@ -2,11 +2,22 @@ import { isAvailable } from "../components/FilterPanel/AvailabilityFilter";
 import { isWithinRadius } from "../components/FilterPanel/RadiusFilter";
 
 const dayjs = require("dayjs");
+const customParseFormat = require("dayjs/plugin/customParseFormat");
```
livgust, line 5: why are we using a mix of requires and imports?

livgust (changes requested):
Sorry for the delay. One nitpicky comment, the rest looks good to me :) thanks!

src/services/appointmentData.service.js
harcod: No reason. I just copy/pasted from dayjs documentation.  It's update in the latest commit!

livgust (approved):

test/fixtures/api/prod-coordinates.json
```diff
@@ -0,0 +1,140 @@
+{
```
tkjones117, line 1: Where is this file used? Is this a locally stored version of test data to use?

tkjones117:
This is very cool! I think it warrants instructions somewhere (in code or in the README) of how to use it, involving steps to 1. update the .env.local var to `false` 2. getting data from AWS 3. rerunning with `npm run test`.
And if that `prod-coordinates` file is used for something, then that too. :)

tkjones117:
Also info on how to undo the changes (get back to fetching real data). I set `REACT_APP_IGNORE_DEVTEST_JSON` back to true, and reran `npm run start`, and I still saw the test data be populated. I needed to actually the delete the test file in order to fetch real data again.

test/fixtures/api/prod-coordinates.json
harcod: We probably don't need this file.  I'll remove it.

.gitignore
```diff
@@ -22,4 +23,3 @@
 npm-debug.log*
 yarn-debug.log*
 yarn-error.log*
-.eslintcache
```
harcod, line 25: This was in the file twice.

harcod:
The env test was not working as expected because "true" is a string (not a boolean) in env files.

Changed the test from the negative of "IGNORE_DEVTEST_JSON" to the affirmative "USE_DEVTEST_JSON".  The production server will not have this set.  Also, there is no longer a need for ".env.test" to set ignore.

@livgust You can also remove the "REACT_APP_IGNORE_DEVTEST_JSON" enviroment variable from the production server.  Now the developer has to explicitly set REACT_APP_**USE**_DEVTEST_JSON" for this test harness to execute.

tkjones117 (approved):
I went through the README instructions to test with fake data (and then undid it by changing the boolean to false, and also by removing the test file). Everything worked!

PR #146, opened 2021-03-13
"Clear Selection" button for filters

### LINKED ISSUE #135: Create "clear filters" button
Refer to spec in issue #130

harcod:
Closes #135 

Clears the filter state back to the default

[X] Only Show Available Appointments
"Blank" ZIP Code
"Any" radius

You still have to press the "Update List" button for the clearing to take effect.


src/components/FilterPanel/index.js
```diff
@@ -50,6 +50,31 @@ const useStyles = makeStyles((theme) => ({
...
+            color="default"
+            className={classes.mobileButton}
+            onClick={(e) => {
+                e.preventDefault();
+                setInProgressFilters({
+                    ...inProgressFilters,
+                    filterByAvailable: true,
+                    filterByZipCode: {
+                        ...inProgressFilters.filterByZipCode,
+                        zipCode: "",
```
livgust, line 66: Should we unset the cookie too? Might be good to move the cookie logic to whenever someone presses either button

livgust (changes requested):
Quick suggestion!

src/components/FilterPanel/index.js
harcod: As it is right now, this button isn't any different from clicking checboxes and clearing numbers.  It doesn't press the "Update List" button.

PR #148, opened 2021-03-16
FAST FIX - Danvers needs invitation code

harcod:
Just filter it out for the moment!


harcod:
No longer need.  This was simultaneously "fixed" in the back-end

PR #155, opened 2021-03-18
Updated "no date-specific" messaging.

### LINKED ISSUE #154: "No date-specific data available"
This was discussed on [Slack](https://macovidvaccines.slack.com/archives/C01N7C2G2KF/p1616074101001100)

Can we look at changing the text for places like CVS that report availability but not dates or quantities?  The message says "No date-specific data available." which is sounds very similar to the "No availability.".   There is a further explanation that you can get by clicking on the (?) icon which states "We were able to determine that there are available appointments, but we can't tell when or how many. Click the sign up button to learn more from the location's website."

"No date-specific data available."  could be replaced with something more positive (without the "No").
Something like "Appointments available.  Click the sign up button to find specific appointment dates". (and get rid of the (?) icon and pop-up)

harcod:
Closes #154 

Updated text and removed the HelpDialog completely since all of the information is displayed on the page.

**NOW**
image

**BEFORE**
image

### DIFF
--- a/src/components/Availability.test.js
+++ b/src/components/Availability.test.js
@@ -15,7 +15,7 @@ it("does not show dates and slot numbers if there aren't any appointments (onlyS
             />
         );
     });
-    expect(screen.queryByText("No availability.")).toBeTruthy();
+    expect(screen.getByTestId("no-availability")).toBeTruthy();
 });
 
 // Test 2
@@ -32,7 +32,7 @@ it("does not show dates and slot numbers if there aren't any appointments (!only
             />
         );
     });
-    expect(screen.queryByText("No availability.")).toBeTruthy();
+    expect(screen.getByTestId("no-availability")).toBeTruthy();
 });
 
 it("shows total slots if slots by date aren't available", async () => {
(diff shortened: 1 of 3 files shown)


src/components/Availability.test.js
```diff
@@ -247,17 +248,18 @@ it("shows total slots across all available days", async () => {
                     appointmentData: {
                         "10/11/2021": {
                             hasAvailability: true,
-                            numberAvailableAppointments: 35,
+                            numberAvailableAppointments: 34,
```
tkjones117, line 251: I assume you tweaked these numbers just to add a little variation.

tkjones117 (approved):
I can't easily verify this since all locations with availability have details at the moment. 
But the code and screenshots look good.

src/components/Availability.test.js
harcod: Yes.  I wanted to make sure that it wasn't counting the first row twice.

iann (approved):
Tested locally and working as expected.

Issue #159, opened 2021-03-19
MoreInformation with one line


harcod:
Per [#ux channel in Slack.](https://macovidvaccines.slack.com/archives/C01N7C2G2KF/p1616112369013700)

If there is only a _map address_ in the **More Information** accordion, then just show that one line and skip the accordion altogether.

> If MORE INFORMATION only displays the address, can we just show the address w/o the accordion? It would take up less space, and could just go where the city/town name is below the site name.

Issue #136, opened 2021-03-09
update cookie with filter selection


livgust:
- Save checkbox state
- Save radio button state
- Update ZIP save - don't save if the ZIP is invalid, and save if it's empty (basically, just update the cookie so it only saves when you press the button instead of when you type)

PR #160, opened 2021-03-20
Show preregistration (massVax) sites

### LINKED ISSUE #156: MassVax sites should appear on our site
Right now we are not showing the existence of the seven MassVax sites (Gilette, Fenway/Hynes, etc.) as there is no **immediate** availability via a _Sign Up_ button

If we flag these sites in the back-end, then we can show them to the user.

1. We will never display the (blue) "no appointments found" message because there is always something to show.
2. MassVax sites show up with ZIP Codes for proximity
3. Removes the (green) "preregistration alert" in favor of this.
4. Promotes **preregistration** to our users

**Here is the approved visual from the [#UX channel on Slack](https://macovidvaccines.slack.com/archives/C01N7C2G2KF/p1616105074009700):**

image

harcod:
Closes #156

Initial Draft of including preregistration buttons for the 7 MassVax sites. (Gilette, Fenway/Hynes, Dartmouth, etc)
There is a [back-end PR](https://github.com/livgust/covid-vaccine-scrapers/pull/161)  that need to be approved that is related to this issue.

**Here is a MassVax Location Card:**
image

**And new Filter UI**
image

To-do:
:heavy_check_mark: 1. update `appointmentData.service.js` to read flag provided by [back-end PR](https://github.com/livgust/covid-vaccine-scrapers/pull/161) (**This will be done by Issue #161** now)
:heavy_check_mark: 2. Add a cookie to save the setting of the checkbox.


harcod:
Closes #159
If the "More Information" accordion only has a street address, then just show the address without the accordion.

harcod:
Closes #136 
Cookies have been refactored into it's own file.  Now saves all "filter" settings in a cookie.  Also the expiration date for the cookie is 1 year (instead of the previous 1 day... oops)

harcod:
I want to move this back to DRAFT status.

harcod:
OK. This time for sure!

src/services/cookie.service.js
```diff
@@ -0,0 +1,64 @@
...
+        encoded
+            .match(/.{1,2}/g)
+            .map((hex) => parseInt(hex, 16))
+            .map(applySaltToChar)
+            .map((charCode) => String.fromCharCode(charCode))
+            .join("");
+};
+
+// To create a cipher
+const cookieSalt = "XDpUY3ti0uEh8wUnWv6t";
```
tkjones117, line 62: I am curious how you came up with this salt?

src/services/cookie.service.js
```diff
@@ -0,0 +1,64 @@
...
+
+const cookies = new Cookies();
+const cookieMaxAge = 365 * 86400;
+const cookieName = "macovidvaccines";
+
+let ourCookie = {};
+
+export function getCookie(key) {
+    const theCookie = cookies.get(cookieName, { doNotParse: true });
+    if (theCookie) {
```
tkjones117, line 11: could you come up with more descriptive names than `theCookie` and `ourCookie`?

tkjones117:
Ran and this works, looks great.

iann:
Something I'm noticing is that the checkbox state isn't being persisted in the cookie. Is this expected?

src/services/cookie.service.js
harcod: random password generator

src/services/cookie.service.js
harcod: Good idea! Changing to `encryptedCookie` and `plaintextCookie`.  These aren't exported, so the most important part about them is their encryption state.

harcod:
> Something I'm noticing is that the checkbox state isn't being persisted in the cookie. Is this expected?

Not expected... it should be persisting **all of the filter settings** now. (The current production version only saves the 5-digit ZIP Code).  I just tried it locally (localhost:3000) and the cookie was being updated when clicking on _Update List_.  Do you have a test case?

iann:
This is more of a Safari issue locally. Confirmed it works locally with Chrome.

iann (approved):
Works/looks good!

livgust (changes requested):
Hey @harcod did you run this final version by UX? It's really annoying to show the mass vax sites by default because the sites with actual availability are peppered between them. I think if we show them, they should be at the bottom unless we're sorting by distance. But in any case, we should definitely run it by UX.

harcod:
> Hey @harcod did you run this final version by UX? It's really annoying to show the mass vax sites by default because the sites with actual availability are peppered between them. I think if we show them, they should be at the bottom unless we're sorting by distance. But in any case, we should definitely run it by UX.

OK. I will get UX to sign off.

harcod:
Closes #161 by checking for the massVax flag in the data file instead of URL matching **color** or **curative**

harcod:
After much discussion in Slack, we decided to change this up quite a bit.

1. Make Preregistration look like a single location.
2. Buttons: [ Preregister ] [ Dismiss this reminder ]
3. Add more text to list all of the locations
4. Sort it.  **First** (I think it is important).  Once they preregister they can dismiss it.
5. The [ Dismiss this reminder ] button saves a cookie so that it never comes back.
6. Remove the "Show mass vaccination locations" in the filter panel

image

src/CovidAppointmentTable.js
```diff
@@ -1,4 +1,5 @@
 /*eslint no-unused-vars: ["error", { "varsIgnorePattern": "[iI]gnored" }]*/
+/*eslint no-unused-vars: ["error", { "varsIgnorePattern": "[iI]gnored" }]*/
```
livgust, line 2: is there a reason this is here twice?

src/services/cookie.service.js
livgust: should this be in the code? should we move it to a secret in the repo or something?

livgust (changes requested):
this looks awesome!! 2 quick comments

src/CovidAppointmentTable.js
harcod: removed.  I probably pressed Ctrl-D in my IDE to see a "diff" and got a "duplicate line" instead.

src/services/cookie.service.js
harcod: moved to secrets

src/services/cookie.service.js
harcod: done

livgust (approved):
Sorry for the delay!

PR #164, opened 2021-03-23
WIP: handle both text or json api response

iann:
Right now the API returns a string as the `body`. This allows for both a string and JSON as the response.


iann:
marking as work in progress for now

harcod:
This looks fine and works with what we have right now.  I would **approve** it, but since it's marked WIP I'll wait for now.

iann:
closing for now

PR #169, opened 2021-04-02
Avoid foo.hasOwnProperty() per eslint:no-prototype-builtins

johnhawkinson:
Instead use the more "fun" and seemingly excessively verbose:
	Object.prototype.hasOwnProperty.call(foo, "prop")

See https://eslint.org/docs/rules/no-prototype-builtins
but basically not all objects need have the methods from
Object.prototype in EMCAScript 5.1, and even if they do they can be
subtly/maliciously overridden.

This tripped me up because I had eslint set to a stricter mode than we do, with `"eslint:recommended"` in my `extends` section of `.eslintrc.js`.

On the other hand, feel free to tell me to "shove it," and that this is needlessly strict and excessively verbose, and we don't need to worry about it

### DIFF
--- a/src/services/appointmentData.service.js
+++ b/src/services/appointmentData.service.js
@@ -149,7 +149,7 @@ export function getAppointmentData() {
             let testData = require("../test/devtest.json");
 
             // If it has 'body', then this looks like something pasted from a browser (View Source)
-            if (testData.hasOwnProperty("body")) {
+            if (Object.prototype.hasOwnProperty.call(testData, "body")) {
                 if (typeof testData.body == "object") {
                     testData = testData.body;
                 } else {


harcod:
I don't have a problem with stricter checking.  Maybe you should add a `.eslintrc.js` to the project so that we are all using the same standard.  Also, add the filename to `.gitignore`

harcod (changes requested):
I don't have a problem with stricter checking. Maybe you should add a `.eslintrc.js` to the project so that we are all using the same standard.

harcod:
Or, can this just be reflected in the "eslintConfig" in `package.json`?

src/services/appointmentData.service.js
```diff
@@ -149,7 +149,7 @@ export function getAppointmentData() {
             let testData = require("../test/devtest.json");
 
             // If it has 'body', then this looks like something pasted from a browser (View Source)
-            if (testData.hasOwnProperty("body")) {
+            if (Object.prototype.hasOwnProperty.call(testData, "body")) {
```
livgust, line 152: maybe dumb question - why can't we do `if ("body" in testData)`?

iann:
> I don't have a problem with stricter checking. Maybe you should add a `.eslintrc.js` to the project so that we are all using the same standard.

I tend to agree. Perhaps update the PR with the `.eslintrc` file so everyone follows the same rules. In the same PR include the fixes required to comply with the new rules (maybe this is the only one?)

PR #178, opened 2021-04-07
Distance Filter Too Restrictive

### LINKED ISSUE #170: "Filters too restrictive" messaging
If the user selects a search radius other than **Any**, then we currently show the "No Appointments Found" (blue) message.

We should alert the user that appointments may be available at other locations outside of their current filter settings.

harcod:
Fixes #170 

image

### DIFF
--- a/src/App.js
+++ b/src/App.js
@@ -114,11 +114,7 @@ function MainComponent() {
             });
     }, []);
 
-    const {
-        filteredData,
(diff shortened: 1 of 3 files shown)


iann (approved):
Excellent UX improvement. I just tested and it works as expected.

hayleyworthman (approved):
The code and UX looks good to me!

PR #182, opened 2021-04-17
MassVax card updated

### LINKED ISSUE #181: MassVax sites expanding
There are several additional MassVax sites being added with more to come.  The messaging should be more generic rather than listing specific locations.

Here is the current UI...

image

harcod:
Fixed #181 

Several additional locations have been added to the Preregistration System recently.  This number currently stands at 7 MassVax sites plus 6 regional collaboratives. The state's site says that "_More locations will continue to be added_."

The text was updated to be more generic so that additional updates to this site will not be necessary as the Commonwealth adds more sites.

In addition, the button was redirected to the state's information page rather than directly sending the user to the registration page.

image

### DIFF
--- a/src/CovidAppointmentTable.js
+++ b/src/CovidAppointmentTable.js
@@ -233,13 +233,10 @@ function MassVaxCard({ className }) {
                     className={classes.massVaxBoxHeader}
                     title={
                         <div className={classes.locationTitle}>
-                            <span>Preregister for Mass Vaccination Sites</span>
+                            <span>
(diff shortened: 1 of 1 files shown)


livgust:
2 things -
- I don't think we need the locations subheader anymore
- the title is now incorrect - prereg isn't just for mass vaccination sites. Maybe just "Preregister" or "Preregistration"?

harcod:
Removed _subheader_ and used copy from the Commonwealth's information page to describe the program.

image

livgust (approved):

PR #184, opened 2021-04-26
Basic React Router installation

harcod:
There is one route "/about" that shows a prototype for an About Us page.  There is no direct navigation to this page in this PR.

"/" shows the current site <App />
"/about" shows the prototype for an About Us page. <About />
every other route redirects to "/" (otherwise we would get a 404 error).

### DIFF
--- a/src/About.js
+++ b/src/About.js
@@ -164,12 +164,10 @@ function AboutContent({ className }) {
                         "mailto:macovidvaccines@gmail.com?subject=site feeedback"
                     }
                 >
-                    send an email send Olivia email
+                    send an email
                 </a>
                 .
             </p>
-
-            {/*            https://www.cnn.com/videos/health/2021/02/08/software-developer-builds-simple-massachusetts-covid-19-vaccine-website-olivia-adams-intv-newday-vpx.cnn*/}
 
             <h3>Technical details</h3>
             <p>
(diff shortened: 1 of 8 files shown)


harcod:
I fixed the "color contrast" Lighthouse error, but now the rel-preconnect fails again.

> uses-rel-preconnect` failure for `maxLength` assertion (Preconnect to required origins: https://web.dev/uses-rel-preconnect/ )

src/About.js
```diff
@@ -0,0 +1,283 @@
...
+                >
+                    making a contribution
+                </a>
+                . If you have questions, or want to get involved,{" "}
+                <a
+                    href={
+                        "mailto:macovidvaccines@gmail.com?subject=site feeedback"
+                    }
+                >
+                    send an email send Olivia email
```
livgust, line 167: I like "send an email"

src/About.js
```diff
@@ -0,0 +1,283 @@
...
+                </a>
+                .
+            </p>
+
+            {/*            https://www.cnn.com/videos/health/2021/02/08/software-developer-builds-simple-massachusetts-covid-19-vaccine-website-olivia-adams-intv-newday-vpx.cnn*/}
```
livgust, line 172: Mind taking this out?

livgust (changes requested):
Thanks for this, it looks good! 2 tiny nits.

livgust (approved):

PR #185, opened 2021-04-27
URL Route for ZIP Codes

harcod:
Allows for specific ZIP Codes to be specified in the URL.  This would enable local government or news websites to publish a link to MACovidVaccines.com with a targeted _ZIP Code_.  Currently, the default when visiting the site with the plain URL ( https://MACovidVaccines.com ) is to show all locations with available appointments **sorted alphabetically** by the location's name.

For example, Worcester, MA could use the direct URL:
https://MACovidVaccines.com/zip/01601

### DIFF
--- a/src/App.js
+++ b/src/App.js
@@ -49,11 +49,11 @@ const useStyles = makeStyles((theme) => ({
     },
 }));
 
-function App() {
+function App({ zipParam }) {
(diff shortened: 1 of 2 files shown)


livgust (approved):
woohoo!

PR #188, opened 2021-04-28
Extracted hard coded strings for translation

harcod:
`translations/translation.en.main.json` is the English translation file.


livgust (approved):
thanks so much Rob!

PR #193, opened 2021-05-11
Consolidate large pharmacy chains

harcod:
Here's the proposal...

1. Remove all CVS, Walgreens, Walmart, and Stop & Shop data from the main feed.  (The Walgreens scraper is currently broken and the Walmart scraper never made it to production)
3. Add a new "Large Pharmacy Chains" card to the top of the listings with links to these 4 major chain pharmacies in Massachusetts.

Since vaccine is widely available at the chains now, we don't really need to be a "store locator" service or even look to see if they have slots anymore.  This will remove lots of locations from our giant list of locations to allow users to find non-pharmacy options in their area.  The Local Boards of Health and MassVax sites now have tens of thousands of slots available.

image

### DIFF
--- a/src/services/appointmentData.service.js
+++ b/src/services/appointmentData.service.js
@@ -93,8 +93,16 @@ function transformData(data) {
     // We are going to show a consolidated "Preregistration" card instead.
     // There is still code to display these sites individually in
     // Availability.js and SignupLink.js if we decide to go that way later on.
+    // ALSO.. Filter out all pharmacies
     mappedData = mappedData.filter((d) => {
-        return !d.isMassVax;
+        return (
+            !d.isMassVax &&
+            d.location &&
+            !d.location.startsWith("CVS (") &&
+            !d.location.startsWith("Stop & Shop") &&
+            !d.location.startsWith("Walgreens") &&
+            !d.location.startsWith("Walmart")
+        );
     });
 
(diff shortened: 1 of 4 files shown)


harcod:
New UI...

image

src/services/appointmentData.service.js
```diff
@@ -93,8 +93,16 @@ function transformData(data) {
     // We are going to show a consolidated "Preregistration" card instead.
     // There is still code to display these sites individually in
     // Availability.js and SignupLink.js if we decide to go that way later on.
+    // ALSO.. Filter out all pharmacies
     mappedData = mappedData.filter((d) => {
-        return !d.isMassVax;
+        return (
```
livgust, line 98: Should we instead:
1) stop scraping those sites & push the change, then
2) delete the parent locations so they aren't passed through to the front-end
?

livgust (approved):
one optional comment; lmk what you think. Approving since it's minor

src/services/appointmentData.service.js
harcod: Here's my thought...

1. Merge this PR now
2. Turn off scrapers ( I did Walgreens already in a PR on the back-end )
3. Change the database retrieval code to not return these 4 sites at all
4. New PR to remove this code that is filtering on the front-end

PR #194, opened 2021-05-12
Updated eligibility for 12+

harcod:
image


johnhawkinson (approved):
Reviewed visually (only); approved.

A narrow targeted diff would be easier for such reviews, and would seem more appropriate to me, a person who tries to make sure that each commit message is longer than the diff it is attached to. I would have put the cleanup in a separate commit (perhaps in the same PR?).

